PR #3, opened 2018-07-09
Add Conduit and Axom, General cleanup

white238:
There were a lot of changes to get it to this point so here are the highlights:

* Added Conduit and Axom builds via their CMake build systems, this allowed using GEOSX's custom hdf5 not the one coming from Axom's spack system
* Locked in all TPL's to mirrored tar/zip files.  This stops TPLs from breaking from underneath you when their repository updates.
* A lot of general clean up in the CMakeList.txt to make everything work the same

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+build-*
+install-*
+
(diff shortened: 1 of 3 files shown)


artv3:
Will these build on CORI?

PR #4, opened 2018-07-12
TPLs on CORI

artv3:
Adjustments to get the TPLS working on CORI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -278,7 +278,7 @@ else()
     set(CHAI_DIR "${CMAKE_INSTALL_PREFIX}/chai")
     set(CHAI_URL "${TPL_MIRROR_DIR}/chai-07_16_2018.tar.gz")
     message(STATUS "Building CHAI found at ${CHAI_URL}")
-    
+
     ExternalProject_Add( chai
                          PREFIX ${PROJECT_BINARY_DIR}/chai
                          URL ${CHAI_URL}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,7 @@ option( RAJA_ENABLE_TBB "" OFF)
 option( ENABLE_CALIPER "" OFF )
 option( ENABLE_TRILINOS "" ON )
 
-set(NUM_PROC "16" CACHE PATH "")
+set(NUM_PROC "8" CACHE PATH "")
 set(TPL_MIRROR_DIR "${CMAKE_SOURCE_DIR}/tpl_mirror")
 set(build_list )
 set(existing_list )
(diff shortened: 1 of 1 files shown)


white238:
Was this a clean repo when you ran config-build/cmake?

artv3:
Yup, I had just cloned the repo and tried to build them.

artv3:
@rrsettgast, perhaps we need to update the CORI host-config? I'm not sure what would be missing.

rrsettgast:
@artv3 what version of cmake is running?

artv3:
@rrsettgast , this is good to go.

CMakeLists.txt
```diff
@@ -273,12 +275,11 @@ endif()
 set(CHAI_DIR "${CMAKE_INSTALL_PREFIX}/chai")
 set(CHAI_URL "${TPL_MIRROR_DIR}/chai-07_16_2018.tar.gz")
 
-if (EXISTS ${CHAI_DIR})
+if ( CHAI_DIR )
```
artv3: @rrsettgast , I think if (EXISTS ${CHAI_DIR}) should be if ( CHAI_DIR ) , otherwise it does not recognize that I've defined my own path to the CHAI directory.

CMakeLists.txt
```diff
@@ -270,13 +271,12 @@ endif()
...
 # CHAI
 ################################
-set(CHAI_DIR "${CMAKE_INSTALL_PREFIX}/chai")
-set(CHAI_URL "${TPL_MIRROR_DIR}/chai-07_16_2018.tar.gz")
-
 if (EXISTS ${CHAI_DIR})
     message(STATUS "Using existing CHAI found at ${CHAI_DIR}")
     list(APPEND existing_list chai )
 else()
+    set(CHAI_DIR "${CMAKE_INSTALL_PREFIX}/chai")
```
white238: This shouldnt be necessary because cache variables are immutable.

white238: Also this is going to cause it not to find an already built chai in the default location.

artv3: If I set ``set(CHAI_DIR "${CMAKE_INSTALL_PREFIX}/chai")`` before ``if (EXISTS ${CHAI_DIR})``  then it jumps to using the locally defined directory rather than the one defined the host-config``set(CHAI_DIR "/global/homes/v/vargas45/Git-Repos/CHAI/build/install" CACHE PATH "" FORCE )`` .  Is my definition in the host-config incorrect?

white238: Okay.. learn something new everyday. Or relearning i guess.

CMake's scoping is janky to say the least.  Local variables take precedence apparently in their directory.  Do this instead:

if (NOT DEFINED CHAI_DIR)
    set(CHAI_DIR "${CMAKE_INSTALL_PREFIX}/chai")
endif()
set(CHAI_URL "${TPL_MIRROR_DIR}/chai-07_16_2018.tar.gz")

PR #22, opened 2018-10-31
Add C/CXX flags to metis and superLU build

artv3:
By passing in the OpenMP flag and std=c99 flags we remove warnings and build errors. Flags are specified from the host config.


artv3:
@rrsettgast , I added a bash script in this PR. Currently I take a step to deinit PAMELA and the PVT package but the step should be removed once they are functional on Cori.

PR #23, opened 2019-01-28
Adding petsc to thirdPartyLibs

hmmorgan:
Build PETSc in thirdPartyLibs


hmmorgan:
Hannah: PETSc includes Trilinos/Hypre/etc in build. Any way to link to existing packages in thirdPartyLibs so that we don't have repeats?

PR #30, opened 2019-05-23
Use LFS instead of tpl_mirror submodule

rrsettgast:
must have git-lfs installed. On LC must select new git-lfs module:
```module load git-lfs/2.6.1```
And then do a
```
git lfs install
git lfs pull
```
after checking out the branch.

Note:
Version 1.4.1 does not work.


rrsettgast:
@corbett5 can this get merged? Does it fix the problems you were having?

corbett5:
@rrsettgast this contains the new Trilinos which I believe is requires some cmake changes in my `feature/corbett/ray` branch. I'll port my changes from that branch here then we can merge this. I'l do that today.

rrsettgast:
@corbett5 I am on "vacation" so would you mind pulling the trigger whenever you are ready?

PR #40, opened 2019-06-21
new interface with blas/lapack

rrsettgast:


CMakeLists.txt
```diff
@@ -103,52 +100,27 @@ else()
...
 
+include( cmake/thirdparty/Find_BLAS.cmake )
+include( cmake/thirdparty/Find_LAPACK.cmake )
 
-################################
-# MKL
-################################
-if (DEFINED ENABLE_MKL AND ENABLE_MKL)
-    message(STATUS "MKL enabled")
+set(BLAS_LIBRARY_DIRS ${BLAS_LIBRARY_DIRS} CACHE PATH "")
```
corbett5: I believe this should be
`set(BLAS_LIBRARY_DIRS ${BLAS_LIBRARY_DIR} CACHE PATH "")`
But we should just keep them `*_LIBRARY_DIR` and deal with the case of multiple directories when we/if we need to.

PR #42, opened 2019-06-25
reenable building on osx

rrsettgast:


CMakeLists.txt
```diff
@@ -40,7 +40,7 @@ option( ENABLE_TRILINOS "" ON )
 
 option( ENABLE_NINJA "" OFF )
 
-set(NUM_PROC "" CACHE STRING "")
```
corbett5: Is there a reason we default it to 16? It seems to me that the default command should be `make -j` or `ninja` and if you want something else you specify it with `-DNUM_PROC=XXX`.

CMakeLists.txt
```diff
@@ -104,19 +106,14 @@ else()
...
-
-    # Check if system is using lib or lib64 as default library installation path
-    get_property(LIB64 GLOBAL PROPERTY FIND_LIBRARY_USE_LIB64_PATHS)
-    set(LIBSUFFIX "")
-    if (${LIB64} STREQUAL TRUE AND ${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86_64")
-        set(LIBSUFFIX 64)
-    endif()
     
-    set(BLAS_LIBRARY_DIRS ${LAPACK_SUITE_DIR}/lib${LIBSUFFIX} CACHE PATH "")
+    set(BLAS_LIBRARY_DIRS ${LAPACK_SUITE_DIR}/${CMAKE_INSTALL_LIBDIR} CACHE PATH "")
```
corbett5: Very nice.

CMakeLists.txt
```diff
@@ -142,8 +143,6 @@ if( ${ENABLE_MKL} )
                                 -D MKL_LIBRARY_NAMES:STRING=${LAPACK_LIBRARY_NAMES}
                                 -D MKL_INCLUDE_DIRS:FILEPATH=${LAPACK_INCLUDE_DIR}
                                 )
-        set(SUPERLU_BLAS_LIBRARIES ${BLAS_LIBRARIES} )
-        set(SUPERLU_LAPACK_LIBRARIES ${LAPACK_LIBRARIES} )
     else()
         set(SUPERLU_BLAS_LIBRARIES "-Wl,-rpath -Wl,${BLAS_LIBRARY_DIRS} -l${BLAS_LIBRARY_NAMES}")
```
corbett5: I don't think the SuperLU BLAS_LIBRARIES are getting set correctly when using the lapack-suite, but I'll fix that next time I rebuild the tpls on lassen (only place we're currently using the lapack-suite).

corbett5: Also we should probably error out if `ENABLE_MKL` is true but `${LAPACK_LIBRARY_NAMES}` doesn't contain "mkl".

PR #44, opened 2019-07-02
Feature/corbett/essl

corbett5:


cmake/thirdparty/FindMathLibraries.cmake
```diff
@@ -0,0 +1,65 @@
...
+        message( FATAL_ERROR "ESSL is enabled but ESSL_INCLUDE_DIRS is not defined." )
+    elseif ( NOT DEFINED ESSL_LIBRARIES )
+        message( FATAL_ERROR "ESSL is enabled but ESSL_LIBRARIES is not defined." )
+    endif()
+
+    set( BLAS_LIBRARIES "${ESSL_LIBRARIES}" CACHE STRING "" FORCE )
+    set( LAPACK_LIBRARIES "${ESSL_LIBRARIES}" CACHE STRING "" FORCE )
+
+else()
+    find_package( BLAS REQUIRED )
```
rrsettgast: if there are multiple BLAS packages on the system, which one does it find? 

If the user wants to specify one through `BLAS_LIBRARIES` will find_package(BLAS) overwrite that specification? 

Also, you need to check that all the required variables were set properly?

corbett5: Fixed. The user can now specify `BLAS_LIBRARIES` and or `LAPACK_LIBRARIES` and the script checks for their existence.

PR #46, opened 2019-07-08
Solutions for superlu and hypre configurations

af1990:
To allow cmake 3.10.2 (default with Ubuntu 18.04) to compile the TPLs, I changed the main `CMakeLists.txt` in two sections:
- superlu_dist: defining `CMAKE_C_COMPILER` as `MPI_C_COMPILER` and `CMAKE_CXX_COMPILER` as `MPI_CXX_COMPILER`. No needs for `MPI_C_COMPILER` and `MPI_CXX_COMPILER`
- hypre: using `.` instead of `source`, to allow cmake to run with `sh` also (not only `bash`)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -632,11 +632,9 @@ ExternalProject_Add( superlu_dist
                                       -DTPL_LAPACK_LIBRARIES:STRING=${LAPACK_LIBRARIES}
                      CMAKE_ARGS -D CMAKE_BUILD_TYPE:STRING=RELEASE
                                 -D XSDK_INDEX_SIZE=64
-                                -D CMAKE_C_COMPILER=${CMAKE_C_COMPILER}
-                                -D CMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
+                                -D CMAKE_C_COMPILER=${MPI_C_COMPILER}
+                                -D CMAKE_CXX_COMPILER=${MPI_CXX_COMPILER}
                                 -D CMAKE_Fortran_COMPILER=${CMAKE_Fortran_COMPILER}
-                                -D MPI_C_COMPILER=${MPI_C_COMPILER}
-                                -D MPI_CXX_COMPILER=${MPI_CXX_COMPILER}
                                 -D CMAKE_C_FLAGS=${C_FLAGS_NO_WARNINGS}
                                 -D CMAKE_C_FLAGS_RELEASE=${CMAKE_C_FLAGS_RELEASE}
                                 -D CMAKE_CXX_FLAGS=${CXX_FLAGS_NO_WARNINGS}
@@ -696,7 +694,7 @@ ExternalProject_Add( hypre
                      INSTALL_DIR ${HYPRE_DIR}
                      DEPENDS superlu_dist
                      BINARY_DIR ${PROJECT_BINARY_DIR}/hypre/src/hypre/src
-                     CONFIGURE_COMMAND cat ${PROJECT_BINARY_DIR}/config_for_geosx && source ${PROJECT_BINARY_DIR}/config_for_geosx
+                     CONFIGURE_COMMAND cat ${PROJECT_BINARY_DIR}/config_for_geosx && . ${PROJECT_BINARY_DIR}/config_for_geosx
                      BUILD_COMMAND make -j ${NUM_PROC} VERBOSE=1
                      INSTALL_COMMAND make install
                    )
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -632,11 +632,9 @@ ExternalProject_Add( superlu_dist
                                       -DTPL_LAPACK_LIBRARIES:STRING=${LAPACK_LIBRARIES}
                      CMAKE_ARGS -D CMAKE_BUILD_TYPE:STRING=RELEASE
                                 -D XSDK_INDEX_SIZE=64
-                                -D CMAKE_C_COMPILER=${CMAKE_C_COMPILER}
-                                -D CMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
+                                -D CMAKE_C_COMPILER=${MPI_C_COMPILER}
+                                -D CMAKE_CXX_COMPILER=${MPI_CXX_COMPILER}
                                 -D CMAKE_Fortran_COMPILER=${CMAKE_Fortran_COMPILER}
-                                -D MPI_C_COMPILER=${MPI_C_COMPILER}
```
rrsettgast: I don't think this should be removed. If anything, `MPI_Fortran_COMPILER` should be set as well.

https://cmake.org/cmake/help/v3.10/module/FindMPI.html

CMakeLists.txt
```diff
@@ -632,11 +632,9 @@ ExternalProject_Add( superlu_dist
                                       -DTPL_LAPACK_LIBRARIES:STRING=${LAPACK_LIBRARIES}
                      CMAKE_ARGS -D CMAKE_BUILD_TYPE:STRING=RELEASE
                                 -D XSDK_INDEX_SIZE=64
-                                -D CMAKE_C_COMPILER=${CMAKE_C_COMPILER}
-                                -D CMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
+                                -D CMAKE_C_COMPILER=${MPI_C_COMPILER}
+                                -D CMAKE_CXX_COMPILER=${MPI_CXX_COMPILER}
```
rrsettgast: These are typically the serial compilers.

CMakeLists.txt
af1990: I know, but nothing changed if I removed them.

af1990: I don't think SuperLU is going to use Fortran.

CMakeLists.txt
af1990: Yes, you're right. But with `serial` compilers, there are missing libraries in the linking phase. It seems that the missing part is provided by the mpi wrappers.

rrsettgast: but it works with a newer version of cmake...correct?

rrsettgast: @white238 please comment on the wonderfulness of cmake FindMPI

af1990: No. The newer cmake is able to use `source` and bypass the hypre issue. The superlu one is still there.

white238: Does SuperLU use FindMPI? Because it doesn't happen by default. If not then you have to use the MPI compilers as the regular ones.

af1990: I don't think so.
From the `superlu` main `CMakeLists.txt`:
```
find_package(MPI)
if(MPI_C_FOUND)
    set(CMAKE_C_FLAGS "${MPI_C_COMPILE_FLAGS} ${CMAKE_C_FLAGS}")
    set(CMAKE_CXX_FLAGS "${MPI_CXX_COMPILE_FLAGS} ${CMAKE_CXX_FLAGS}")
    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${MPI_C_LINK_FLAGS}" )
endif()
```

rrsettgast: `find_package(MPI)` is `FindMPI`.
@corbett5 and yet this works on our systems?

corbett5: What currently exists works on both Quartz and Lassen. I was suggesting reverting to both `CMAKE_C_COMPILER` and `MPI_C_COMPILER` being `${MPI_C_COMPILER}` (same for CXX).

rrsettgast: ok then. just make sure it works on quartz and lassen with cmake 3.14.xx.

corbett5: Can't we still use 3.9?

rrsettgast: I don't know. But I think the cmake treatment of CUDA has changed dramatically since 3.9. If I were a betting man, my money would be on nyet.

corbett5: Ok. Well I suggest we define both `CMAKE_C_COMPILER` and `MPI_C_COMPILER` then merge this and I'll figure things out on quartz/lassen next time I update the tpls. I'd prefer not to attempt to rebuild them these next two weeks.

rrsettgast: I agree...

af1990: Well. I'm going to reintroduce `MPI_C_COMPILER` now.

PR #52, opened 2019-08-29
Reduce Trilinos lib list

klevzoff:
This removes Stokhos, which takes an enormous amount of time and memory to compile, and a few other unused packages (Sacado, PAMGEN, Intrepid, Shards). Forgot to time the builds, but it's significantly faster and less memory hungry.

Also passes same `CMAKE_BUILD_TYPE` to Trilinos as in top-level project, to enable actual debug build of Trilinos. I happen to use this quite often when debugging DofManager and other LAI parts. Most people will link debug build of GEOSX against release TPLs, so they won't see any difference.

GEOSX builds and passes all tests against this update. So I don't think I removed anything useful.


TotoGaz:
Looks consistent with the current build system.

A question: why do we specify all the targets by hand in the script and not a simple 'make' that considers all the targets.
And if there are useless targets, why don't we simply remove them? (Git will allow a smooth rollback if needed).

klevzoff:
> A question: why do we specify all the targets by hand in the script and not a simple 'make' that considers all the targets.
> And if there are useless targets, why don't we simply remove them? (Git will allow a smooth rollback if needed).

I think this was specifically to disable building Trilinos on Travis, because it was too massive/took too long. On a local machine, I always do just `make`.

PR #110, opened 2020-05-23
Update doxygen to 1.8.20

klevzoff:
This version fixes some of the false warnings we're having with 1.8.13, but does have some unresolved problems of its own. The biggest one is that it doesn't recognize zero-parameter methods overriding a template base interface as such, which affects our LAI matrix/vector classes (the problem is reported here: https://github.com/doxygen/doxygen/issues/7304). Nevertheless, it seems like a worthwhile upgrade.

https://github.com/GEOSX/GEOSX/pull/1181


rrsettgast:
@TotoGaz Can you take a look at the maxosx test failure. It seems to fail the google auth.

PR #132, opened 2020-10-01
Hypre AMG on GPU

rrsettgast:


rrsettgast:
@bmhan12 Can you help me out with this PR? I am getting an error on the `gcc-cuda` build:

```_hypre_utilities.hpp:27:10: fatal error: cublas_v2.h: No such file or directory
3184 #include <cublas_v2.h>
3185
```

This is because the file doesn't exist in the cuda path. It appears to exist in the 10.2 path, but I am don't understand why that is there. I think there was some tricks you had to play to get this image working. I think it would be better to have you fix it. 

Thanks
Randy

bmhan12:
@rrsettgast Interesting... I'll look into it.

bmhan12:
@rrsettgast  The Docker image builds locally for me but not on Travis. Looking into a fix.

rrsettgast:
@corbett5 Can you review

docker/gcc-cuda/Dockerfile
```diff
@@ -118,6 +121,10 @@ ARG CUDA_TOOLKIT_ROOT_DIR=/usr/local/cuda
 ARG CUDA_ARCH=sm_70
 ARG CMAKE_CUDA_COMPILER=${CUDA_TOOLKIT_ROOT_DIR}/bin/nvcc
 
+ENV HYPRE_CUDA_SM=70
+ENV CUDA_HOME=${CUDA_TOOLKIT_ROOT_DIR}
+#RUN ln -s ln -s /usr/include/cublas_v2.h /usr/local/cuda/include/cublas_v2.h
+
```
rrsettgast, line 127:
```suggestion
```

docker/gcc-cuda/Dockerfile
```diff
@@ -57,6 +59,7 @@ RUN yum install -y \
 
 ENV LIBRARY_PATH /usr/local/cuda/lib64/stubs
 
+RUN yum install -y yum-plugin-versionlock && yum versionlock libcublas10
```
rrsettgast, line 62:
```suggestion
```

CMakeLists.txt
```diff
@@ -83,6 +90,33 @@ message(STATUS "Build command = ${TPL_BUILD_COMMAND}")
 include(GNUInstallDirs)
 include( cmake/thirdparty/FindMathLibraries.cmake )
 
+
+set( GEOSX_LOCALINDEX_TYPE "std::ptrdiff_t" CACHE STRING "" )
+set( GEOSX_GLOBALINDEX_TYPE "long long int" CACHE STRING "" )
+
+if( GEOSX_LOCALINDEX_TYPE STREQUAL "int" )
```
corbett5, line 97: What is the point of this conversion?

corbett5 (approved):

CMakeLists.txt
rrsettgast: We have to support `int` index types for hypre on the gpu....so fun.

corbett5: Yeah I get that, I'm just wondering what the point of the `GEOSX_LOCALINDEX_TYPE_FLAG` is, the conversion seems unnecessary and complicated. Like how am I supposed to know what `3` means. Why not just keep the string?

PR #135, opened 2020-10-16
Correct inconsistency between pip and python

TotoGaz:
`pip3` was installing packages for a version of python we were not using.
We force `pip3` to point to the actual `python3` in the `PATH` so we are consistent again.


TotoGaz:
@rrsettgast You can merge this PR and rebase your work on it, it should be OK 🤞

rrsettgast (approved):
@TotoGaz Thanks!

PR #145, opened 2021-02-09
Provide MPI fortran compiler to Trilinos

TotoGaz:
The Fortran compiler was not the MPI one and I actually experienced 
a building error in some configuration because of this discrepancy.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -675,8 +675,8 @@ if (ENABLE_TRILINOS)
                                     -D CMAKE_CXX_COMPILER:PATH=${TRILINOS_CXX_COMPILER}
                                     -D CMAKE_CXX_FLAGS:STRING=${TRILINOS_CXX_FLAGS}
                                     -D CMAKE_CXX_FLAGS_RELEASE:STRING=${CMAKE_CXX_FLAGS_RELEASE}
-                                    -D CMAKE_Fortran_COMPILER:PATH=${CMAKE_Fortran_COMPILER}
-                                    -D CMAKE_Fortran_FLAGS_RELEASE:STRING=${CMAKE_Fortran_FLAGS_RELEASE}
+                                    -D CMAKE_Fortran_COMPILER:PATH=${TRILINOS_Fortran_COMPILER}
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -675,8 +675,8 @@ if (ENABLE_TRILINOS)
                                     -D CMAKE_CXX_COMPILER:PATH=${TRILINOS_CXX_COMPILER}
                                     -D CMAKE_CXX_FLAGS:STRING=${TRILINOS_CXX_FLAGS}
                                     -D CMAKE_CXX_FLAGS_RELEASE:STRING=${CMAKE_CXX_FLAGS_RELEASE}
-                                    -D CMAKE_Fortran_COMPILER:PATH=${CMAKE_Fortran_COMPILER}
-                                    -D CMAKE_Fortran_FLAGS_RELEASE:STRING=${CMAKE_Fortran_FLAGS_RELEASE}
+                                    -D CMAKE_Fortran_COMPILER:PATH=${TRILINOS_Fortran_COMPILER}
+                                    -D CMAKE_Fortran_FLAGS_RELEASE:STRING=${TRILINOS_Fortran_COMPILER}
```
corbett5, line 679: This can't be right. Although since it didn't produce an error I guess Trilinos isn't using them?

corbett5 (approved):

CMakeLists.txt
TotoGaz: I'm not sure about `trilinos` but I found myself in a context where it was not working without this patch.
So it may have an influence. 🤷

corbett5: My comment was about `CMAKE_Fortran_FLAGS_RELEASE:STRING=${TRILINOS_Fortran_COMPILER}`, why did you change it from `CMAKE_Fortran_FLAGS_RELEASE` to `TRILINOS_Fortran_COMPILER`?

TotoGaz: Oh, it's a mistake and I misunderstood your comment. Since you did approve the PR I merged it.
Indeed, it's some strange behavior here.
I will correct right now.

PR #150, opened 2021-02-16
Add MOAB (Mesh-Oriented datABase) as a mandatory dependency

TotoGaz:
To be noted: 

- It passes the tests.
- Untested without `mpi`.
- I believe it builds `static` by default. Do we need the `shared` object libraries?
- No `scotch`/`ptscotch`/`zoltan` mesh partitionning.
- The `vtk` feature requires `X11`.
- It builds with `netcdf`.

First shot on smooth `gcc`/`openmpi` on `debian`, no `CUDA`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -799,7 +799,7 @@ ExternalProject_Add( moab
                      PREFIX ${PROJECT_BINARY_DIR}/moab
                      INSTALL_DIR ${MOAB_DIR}
                      DEPENDS ${MOAB_DEPENDENCIES}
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -773,6 +772,45 @@ ExternalProject_Add( metis
 
 list(APPEND build_list metis )
 
+################################
+# MOAB (Mesh-Oriented datABase) is a mandatory dependency
+################################
+
+# MOAB is a mandatory dependency
```
corbett5, line 779: If this is going to be mandatory you will need to add it to our Spack build.

corbett5: I really don't want to do it, and other people need to start using it.

PR #155, opened 2021-04-14
Add Legacy IO support while compiling vtk

AntoineMazuyer:
https://github.com/GEOSX/GEOSX/pull/1453


AntoineMazuyer:
I have an issue with the centos build : 

```
In file included from /tmp/build/vtk/src/vtk/Parallel/DIY/vtkDIYUtilities.h:32:
/tmp/build/vtk/src/vtk/ThirdParty/diy2/vtkdiy2/include/vtkdiy2/serialization.hpp:97:24: error: no member named 'is_trivially_copyable' in namespace 'std'
    static_assert(std::is_trivially_copyable<T>::value, "Default serialization works only for trivially copyable types");
                  ~~~~~^
[0m[91m/tmp/build/vtk/src/vtk/ThirdParty/diy2/vtkdiy2/include/vtkdiy2/serialization.hpp:97:46: error: 'T' does not refer to a value
[0m[91m    static_assert(std::is_trivially_copyable<T>::value, "Default serialization works only for trivially copyable types");
                                             ^
/tmp/build/vtk/src/vtk/ThirdParty/diy2/vtkdiy2/include/vtkdiy2/serialization.hpp:92:18: note: declared here
  template<class T>
                 ^
```

When I google this issue, I find that it can be related to a partial implementation of C++11, but it seems related to GCC so I really don't know what to do with that.

klevzoff:
Clang picks up a default c++ standard library, which on that centos version is from gcc-4.8.5. That version claims c++11 conformance, but is actually missing some `type_traits`. I find the practice of clang not shipping with its own standard library _by  default_ extremely weird. In any case, we need to update the docker image, and install either a newer `libstdc++` or `libc++`.

AntoineMazuyer:
@rrsettgast do you know if the clang 9 centos build is still used in LLNL ?

AntoineMazuyer:
@bmhan12 do you have any idea how to overcome this issue ?

bmhan12:
@AntoineMazuyer Would it be possible to patch out the check itself (I'm assuming it's this check [here](https://github.com/Kitware/VTK/blob/6a9c565da01bcd6295d0bcbb66c0a9d0d2eaa69e/ThirdParty/diy2/vtkdiy2/include/vtkdiy2/serialization.hpp#L95))? Based on this [stackoverflow discussion](https://stackoverflow.com/questions/25123458/is-trivially-copyable-is-not-a-member-of-std) and searching vtk, this instance of `is_trivially_copyable` is the only use of partially implemented type traits.

AntoineMazuyer:
@bmhan12 thanks, I patched the lines (commit #9c7337f). I think it's now ready to be merged :)

klevzoff (approved):

PR #156, opened 2021-04-19
Cleanup hypre build

klevzoff:
This PR does the following:
 * Enclose in `ENABLE_HYPRE`
 * Pass proper debug flags when building a Debug configuration (also for a few other packages)
 * Use correct variable for hypre-GPU (`ENABLE_CUDA` -> `ENABLE_HYPRE_CUDA`)
 * Pass `CUDA_HOME` and `HYPRE_CUDA_SM` variables as appropriate
 * Simplify config logic using a single write command (as opposed to MPI and non-MPI cases previously)

The diff may be a bit weird because of indentation changes.

TPL tag will be updated via https://github.com/GEOSX/GEOSX/pull/1397


CMakeLists.txt
```diff
@@ -44,15 +44,10 @@ include(ExternalProject)
 option( BUILD_SHARED_LIBS "" OFF )
 
 option( ENABLE_CALIPER "" ON )
+OPTION( ENABLE_HYPRE "" ON )
```
rrsettgast, line 47: lowercase?

CMakeLists.txt
```diff
@@ -75,6 +70,8 @@ string(REPLACE "-Wextra" "" CXX_FLAGS_NO_WARNINGS ${CXX_FLAGS_NO_WARNINGS})
 set(CXX_FLAGS_NO_WARNINGS "${CXX_FLAGS_NO_WARNINGS} -w")
 string(STRIP ${CXX_FLAGS_NO_WARNINGS} CXX_FLAGS_NO_WARNINGS)
 
+string(TOUPPER ${CMAKE_BUILD_TYPE} BUILD_TYPE_UPPER)
```
rrsettgast, line 73: This is only used for hypre? Should it be guarded?

CMakeLists.txt
```diff
@@ -827,119 +822,94 @@ endif()
...
   ./configure \
-  CC=${MPI_C_COMPILER} \
+  CC=${HYPRE_C_COMPILER} \
   CFLAGS=\"${HYPRE_C_FLAGS}\" \
   CXX=${HYPRE_CXX_COMPILER} \
   CXXFLAGS=\"${HYPRE_CXX_FLAGS}\" \
   FC=${HYPRE_Fortran_COMPILER} \
   FCFLAGS=\"${HYPRE_Fortran_FLAGS}\" \
+  CUDA_HOME=\"${CUDA_TOOLKIT_ROOT_DIR}\" \
+  HYPRE_CUDA_SM=\"${HYPRE_CUDA_SM}\" \
```
rrsettgast, line 888: Does this make it so we don't have to set the environment variables?

CMakeLists.txt
```diff
@@ -827,119 +822,94 @@ endif()
...
     set( HYPRE_C_COMPILER ${MPI_C_COMPILER} )
     set( HYPRE_CXX_COMPILER ${MPI_CXX_COMPILER} )
     set( HYPRE_Fortran_COMPILER ${MPI_Fortran_COMPILER} )
-    if ( ENABLE_CALIPER_HYPRE )
-        set(HYPRE_DEPENDS superlu_dist caliper)
-    else()
-        set(HYPRE_DEPENDS superlu_dist)
+    set( HYPRE_SUPERLU_LIBS "${SUPERLU_DIR}/lib/libsuperlu_dist.a ${PARMETIS_DIR}/lib/libparmetis.a ${METIS_DIR}/lib/libmetis.a -lstdc++" )
+    if( ENABLE_OPENMP )
+      set( HYPRE_SUPERLU_LIBS "${HYPRE_SUPERLU_LIBS} ${OpenMP_C_FLAGS}" ) # Sergey: why is this needed?
```
rrsettgast, line 845: Does it fail without this??

rrsettgast (approved):

CMakeLists.txt
klevzoff: I... don't know how this happened :)

CMakeLists.txt
klevzoff: I'm not sure, to be honest. I can build Hypre with CUDA this way without env vars. And I'm still having device linking issues on the GEOSX side, regardless of how hypre was built and whether or not env vars are set. I added a workaround by forcing device linking on `libgeosx_core` [here](https://github.com/GEOSX/GEOSX/blob/b9df14663d1d188e3ea554086d65bea64164b9c2/src/coreComponents/CMakeLists.txt#L98-L101), but it doesn't work when object libraries are enabled, since we're not building `libgeosx_core` in that case and I'd have to use a similar trick to force device linking on every single executable target - I did this only for `geosx` but not any of the test binaries. I don't think my linking issues are related to these hypre build flags, but because I can't answer with 100% certainty yet.

CMakeLists.txt
klevzoff: I haven't tested, but it was set this way before, so I assume yes, it would fail without. It just made me wonder why `-fopenmp` needs to be added to `--with-dsuperlu-lib=` list of libraries. Maybe @castelletto1 knows.

CMakeLists.txt
klevzoff: Right now, yes. In general, I'd expect it to be useful anytime we want to pass debug flags to a non-CMake subproject, though I haven't considered changing other TPLs yet. I wouldn't guard it since there is nothing inherently hypre-specific about it.

klevzoff: Btw this whole PR was motivated by me needing to build hypre with `-g -O0` so I could debug a crash. I've occasionally found it useful to build GEOSX aganst debug versions of other TPLs too, but we haven't systematically enabled it. This PR gets maybe halfway there.

CMakeLists.txt
castelletto1: Honestly I don't know. It was introduced [here](https://github.com/GEOSX/thirdPartyLibs/commit/28e7581ff451e6d4b1b8b4212343b248463dc8f3#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a). @corbett5 any thoughts?

corbett5: Yeah superlu depends on OpenMP.
https://github.com/GEOSX/thirdPartyLibs/blob/349fb96c34d8df7098c05710ed956ba59f88b5c8/CMakeLists.txt#L820

PR #158, opened 2021-07-14
Add CMP0025 patch for CHAI

klevzoff:
Resolves https://github.com/GEOSX/thirdPartyLibs/issues/157


klevzoff:
@rrsettgast I'm keeping the `PATCH_COMMAND`-based solution instead of updating whole of CHAI because I don't feel like testing a new release (let alone an unreleased development snapshot) right now. Whoever goes on to update and test CHAI next should just remove the patch.

rrsettgast (approved):

PR #159, opened 2021-07-30
Remove unused TPLs

klevzoff:
This PR removes the following TPLs not used in GEOSX:
  * `Astyle`
  * `Axom`
  * `fparser`
  * `silo-4.10.2` (there is still a `4.10.3` that we're building)

No matter how fast they are to build, it just doesn't make much sense to keep them around, as they've remained unused for years. If anyone had plans for any of these dependencies to be used in the future, let me know.

A GEOSX PR will follow with minimal changes (TPL tag update and removing mention of Axom from the Build Guide).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -141,26 +141,6 @@ ExternalProject_Add( uncrustify
 
 list(APPEND build_list uncrustify )
 
-################################
-# ASTYLE
-################################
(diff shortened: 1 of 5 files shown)


rrsettgast (approved):

rrsettgast:
Thanks!!

PR #161, opened 2021-09-11
Add {fmt}

klevzoff:
Related to https://github.com/GEOSX/GEOSX/pull/1568

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1023,6 +1023,31 @@ if(ENABLE_VTK)
 endif(ENABLE_VTK)
 
 ################################
+# FMT
+################################
+set(FMT_DIR "${CMAKE_INSTALL_PREFIX}/fmt")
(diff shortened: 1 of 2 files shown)


klevzoff:
@rrsettgast this is going to require a TPL update on LC once merged. The good thing is `{fmt}` itself takes only a few seconds to build, so if you just pull into an existing build tree it'll be super quick.

klevzoff:
@rrsettgast @CusiniM @francoishamon Could one of you gentlemen or someone else please build this branch on LC and push updated host-configs to https://github.com/GEOSX/GEOSX/pull/1568 if needed (I don't know exactly how the TPL process works on your side). Whenever you have time, but I'd like that PR to go in sooner rather than later. The build should be super quick if you're resuming from an existing build tree, since only one tiny lib is added.

CusiniM (approved):

PR #163, opened 2021-09-29
Add explicit visibility flags to override {fmt} defaults

klevzoff:
This PR explicitly overrides options in `{fmt}`'s CMake that add `-fvisibility=hidden` compiler flag and result in linker warnings on MacOS.

@herve-gross @CusiniM Could one of you test this branch on your Macs please? (I know I can do it on Travis, but I don't have a GEOSX PR open atm). Do something like
```sh
cd thirdPartyLibs/build-darwin-clang-release
git fetch origin
git checkout bugfix/klevzoff/fmt-visibility
rm -rf fmt
cmake ..
make fmt
```
and then build GEOSX and see if the `ld` warning reappears.


CusiniM (approved):
Yep, it seems to work!

PR #165, opened 2021-10-05
Adding a Pecan CPU target alongside the GPU one

TotoGaz:
Some users are now developing on Pecan and not only launching computations.
They would greatly benefit from both `CPU` and `GPU` pre-compiled TPLs.


bmhan12 (approved):
LGTM!

PR #169, opened 2021-10-29
Move to ubuntu 20.04 and gcc 9 & 10

TotoGaz:
* Change ubuntu18.04 to ubuntu20.10.
* Change compilers from gcc8 to gcc 9 & 10.
* Improve description of targets

The parametrized `Dockerfile` is actually some kind of revert of https://github.com/GEOSX/thirdPartyLibs/commit/db5737360d64a8993ef9cf010257848c9d091610 which we've been using successfully for quite some time before removing support of gcc 7.

I still need to rename `gcc-ubuntu18.04` into `gcc-ubuntu`.
I'll do it after the review (I would not like to make the review more complicated than what it needs to be).

Related to https://github.com/GEOSX/GEOSX/pull/1700


klevzoff:
@TotoGaz could you try with 20.04? 
1) It's an LTS release (so likely in much wider use), while 20.10 is already superseded by other bleeding-edge releases and out of support
2) 20.10 ships with `make` 4.3 which is causing the `\#` trouble for older Trilinos: https://lwn.net/Articles/810071/ (as you've already noticed), while 20.04 uses `make` 4.2.1, so you wouldn't be forced to upgrade Trilinos at the same time with OS images.

bmhan12 (approved):
LGTM!
One change is rename `docker/gcc-ubuntu1804/Dockerfile` to `docker/gcc-ubuntu2004/Dockerfile`, or maybe just `gcc-ubuntu`.

TotoGaz:
> LGTM! One change is rename `docker/gcc-ubuntu1804/Dockerfile` to `docker/gcc-ubuntu2004/Dockerfile`, or maybe just `gcc-ubuntu`.

Yes, I will replace with `gcc-ubuntu` with no date.

PR #172, opened 2022-01-21
buildrootdir/installrootdir options in config-build.py

XL64:
Adds a build root dir / build path distinction.
The first option will keep the auto-generated folder name. The latter is unchanged.
Two options overlap somehow: the script fails if both options are provided.

(I have the same change comming to GEOSX scripts)

### DIFF
--- a/scripts/config-build.py
+++ b/scripts/config-build.py
@@ -26,16 +26,34 @@ def extract_cmake_location(file_path):
 
 parser = argparse.ArgumentParser(description="Configure cmake build.")
 
-parser.add_argument("-bp",
-                    "--buildpath",
-                    type=str,
-                    default="",
-                    help="specify path for build directory.  If not specified, will create in current directory.")
+group = parser.add_mutually_exclusive_group()
+group.add_argument("-bp",
+                   "--buildpath",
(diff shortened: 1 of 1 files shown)


XL64:
See https://github.com/GEOSX/GEOSX/pull/1741

TotoGaz:
@XL64 is `config-build.py` up to date w.r.t. the GEOSX repo?

Also, your builds are failing when uploading to `dockerhub`. I think it's because you are using a fork, and therefore you can't access the credentials.
Would you mind opening the PR directly from the `GEOSX/thirdPartyLibs` repository? You can even do it from the UI, no need to do another clone.
Sorry for the inconvenience.

XL64:
Ok i'll do so

XL64:
Didn't found how to do it via the UI, I created a new one, is it ok ? shall I delete this one ?

TotoGaz:
> Didn't found how to do it via the UI, I created a new one, is it ok ? shall I delete this one ?

I meant that you can modify a file and create a branch directly from the UI.
Which could have been convenient for your case.

I'l close this PR, don't worry.

PR #175, opened 2022-01-28
buildrootdir/installrootdir options in config-build.py

XL64:
Add possibility to change build/install root directory while keeping automatic build/install directory naming following configuration file

config-build remove tpl option, forbid -ip with -ir and -br withj -br


TotoGaz (approved):

TotoGaz:
There are some discrepancies between this version and https://github.com/GEOSX/GEOSX/pull/1741 but it's not your patch!

XL64:
Indeed, i didn't know if all should match nor how to merge them.

PR #178, opened 2022-03-08
Update VTK version and patch a duplication bug

klevzoff:
This PR updates VTK to latest and applies a patch that fixes a problem with duplicated mesh nodes resulting from merging mesh partitions assigned to the same rank (which only occurs when number of ranks is not a power of 2). There is also a fix for this in upstream VTK (not in any release version yet), but that fix caused some other issues that need more investigation. I propose that we move ahead with the patch and re-evaluate later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ geosx_osx_build: &geosx_osx_build
   - git lfs pull
   script:
   - GEOSX_TPL_DIR=/usr/local/GEOSX/GEOSX_TPL && sudo mkdir -p -m a=rwx ${GEOSX_TPL_DIR}/..
-  - python scripts/config-build.py -hc ${TRAVIS_BUILD_DIR}/host-configs/darwin-clang.cmake -bt Release -DNUM_PROC=$(nproc) -DGEOSXTPL_ENABLE_DOXYGEN:BOOL=OFF -ip ${GEOSX_TPL_DIR}
+  - python scripts/config-build.py -hc ${TRAVIS_BUILD_DIR}/host-configs/darwin-clang.cmake -bt Release -ip ${GEOSX_TPL_DIR} -DNUM_PROC=$(nproc) -DGEOSXTPL_ENABLE_DOXYGEN:BOOL=OFF -DENABLE_VTK:BOOL=OFF
   - cd build-darwin-clang-release
   - make
   # The deployment phase is done here because the `deploy` stage of travis does not work for PR.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -974,7 +974,7 @@ endif()
 ################################
 if( ENABLE_VTK )
     set( VTK_DIR "${CMAKE_INSTALL_PREFIX}/vtk" )
-    set( VTK_URL "${TPL_MIRROR_DIR}/VTK-9.0.3.tar.gz" )
+    set( VTK_URL "${TPL_MIRROR_DIR}/VTK-9.1.0.tar.gz" )
 
     message( STATUS "Building VTK found at ${VTK_URL}" )
 
@@ -1002,7 +1002,9 @@ endif( ENABLE_MPI )
                          PREFIX ${PROJECT_BINARY_DIR}/vtk
                          URL ${VTK_URL}
                          INSTALL_DIR ${VTK_DIR}
-                         PATCH_COMMAND patch --directory=${PROJECT_BINARY_DIR}/vtk/src/vtk/ThirdParty/diy2/vtkdiy2/include/vtkdiy2 < ${TPL_MIRROR_DIR}/VTK-9.0.3-disable_traits.patch
+                         PATCH_COMMAND patch -p0 < ${TPL_MIRROR_DIR}/VTK-9.1.0-disable_traits.patch
+                               COMMAND patch -p0 < ${TPL_MIRROR_DIR}/VTK-9.1.0-duplicate-points-fix.patch
+                               COMMAND patch -p0 < ${TPL_MIRROR_DIR}/VTK-9.1.0-cmake-fix.patch
                          BUILD_COMMAND ${TPL_BUILD_COMMAND}
                          INSTALL_COMMAND "${TPL_INSTALL_COMMAND}"
                          CMAKE_GENERATOR ${TPL_GENERATOR}
--- a/tplMirror/VTK-9.0.3.tar.gz
+++ b/tplMirror/VTK-9.0.3.tar.gz
@@ -1,3 +1,0 @@
-version https://git-lfs.github.com/spec/v1
-oid sha256:bc3eb9625b2b8dbfecb6052a2ab091fc91405de4333b0ec68f3323815154ed8a
-size 34684378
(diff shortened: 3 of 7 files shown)


tplMirror/VTK-9.1.0-DuplicatePointsFix.patch
```diff
@@ -0,0 +1,10 @@
+--- Filters/ParallelDIY2/vtkRedistributeDataSetFilter.cxx	2022-03-08 00:35:31.589443296 -0800
++++ Filters/ParallelDIY2/vtkRedistributeDataSetFilter.cxx.patched	2022-03-08 00:36:27.214917422 -0800
+@@ -518,6 +518,7 @@
+   {
+     assert(vtkUnstructuredGrid::SafeDownCast(outputDO) != nullptr);
+     vtkNew<vtkAppendFilter> appender;
++    appender->MergePointsOn();
```
TotoGaz, line 7: I assume this is related to https://gitlab.kitware.com/vtk/vtk/-/commit/7016a55484b7a5429d1e74db0c78805607d7d77e ?
But the original patch looks broader? Why do you only keep this line?

klevzoff: Yes, exactly that :smile: 
That whole commit caused some other issues in my tests, in particular one of cell blocks in `cube.vtu` test went entirely missing when running on 3 ranks. I didn't have time for a full investigation or making a minimal reproducer to report back to VTK yet. But I think it's because of the deleted lines in that commit. This is a temporary solution, but I need it in order to lift the power-of-two restriction of `VTKMeshGenerator` (I will push the relevant changes to that PR soon).

klevzoff:
Oh no, the new VTK version relies on some CMake 3.16 functionality :fearful: This might be a no-go b/c of our various platforms, and I might need to find a way to apply the patch on top of 9.0.3 (which is much more difficult). Or should I apply another patch to patch their use of CMake? :suspect:

TotoGaz:
> Oh no, the new VTK version relies on some CMake 3.16 functionality 

Let's just upgrade our `cmake` requirements?
On my company's side, I expect this to be quite loose...

tplMirror/VTK-9.1.0-DuplicatePointsFix.patch
TotoGaz: > in particular one of cell blocks in `cube.vtu` test went entirely missing when running on 3 ranks

Are you saying that we got a successful `vtk` test ? 🏄 🎉 
Jokes aside, does `vtk 9.1` remove this power of two limitation?

klevzoff:
> Let's just upgrade our `cmake` requirements?
> On my company's side, I expect this to be quite loose...

I'm not opposed, but I'm not sure about LC?
VTK actually claims to support 3.12 and up, so I think that [change](https://gitlab.kitware.com/vtk/vtk/-/commit/6343855027c54bed98183d5b4001f02eb1c1b705) in their scripting that causes an error with our 3.14 was accidental/untested. Their entire CI seems to be running on 3.19 (I just had a quick glance, so maybe not an accurate statement).

tplMirror/VTK-9.1.0-DuplicatePointsFix.patch
klevzoff: > Jokes aside, does `vtk 9.1` remove this power of two limitation?

`vtkRedistributeDataSetFilter` never formally had this limitation, but it's API isn't particularly forthcoming to extract the assignment of partition bounding boxes to ranks, which is required to compute MPI neighbors on our side. I opened a [discussion](https://discourse.vtk.org/t/vtkredistributedatasetfilter-cuts-to-mpi-ranks/7927), but the solution proposed to me was a bit too complicated. I ended up implementing a piece of code that basically duplicates how partition/rank assignment is done internally in VTK. When testing it, I also found some bugs in the filter (upon merging partitions on the same rank common nodes were duplicated), which were fixed by a combination of patches, some of which made it into 9.1.0 and some didn't (the one you referenced above).

So to be precise, a combination of 9.1.0, my custom patch and a bit of extra code in `VTKMeshGenerator` (just pushed to https://github.com/GEOSX/GEOSX/pull/1791) does indeed remove the limitation. I just ran an integrated test case on every MPI size from 1 to 8.

klevzoff:
> Oh no, the new VTK version relies on some CMake 3.16 functionality  This might be a no-go b/c of our various platforms, and I might need to find a way to apply the patch on top of 9.0.3 (which is much more difficult). Or should I apply another patch to patch their use of CMake? :suspect:

I ended up:
  * Submitting an [issue](https://gitlab.kitware.com/vtk/vtk/-/issues/18487) in VTK
  * Creating another patch to remove the offending line from VTK's CMake

klevzoff:
After several attempts, I'm unable to pass the Mac OSX build because of timeouts :(

Build logs for VTK on Travis are pretty much identical between the 9.0.3 and 9.1.0 versions. In particular, both builds produce exactly 5903 lines of output, making it extremely unlikely that the new version is by-default including more files/modules into the build. Of course some of the existing files might take longer to build due to code changes, but the difference between last successful build (2h24m) and current timeouts (around 2h40m) is too high IMO to be explained by it. Maybe it's just the usual variation in Travis VMs performance?

TotoGaz:
Sometimes you get really long time for `brew`. What we do with `brew` is not great so it may be worth having a look at that.
I could see a 20/25 minutes duration for `brew`, which, while long, does not look crazy _w.r.t._ other builds (I think).
Maybe upgrading to a more recent hash of brew would help get better caches?

TotoGaz:
> I thought we had it at 3 hrs?

I think so.

klevzoff:
Anyway, 2 cores is not an appropriate number to build extensive large libraries like Trilinos or VTK in 2022. If OSX VMs cannot be buffed, I would vote to abandon it altogether, or look for a different CI service.

TotoGaz:
As always, does any of us have time for any migration? Plus it's not so easy to get some OSX machines in the cloud.

klevzoff:
@TotoGaz @rrsettgast could you approve the PR for the merge please? Feel free to pull the trigger on it immediately as well.

.travis.yml
```diff
@@ -24,7 +24,7 @@ geosx_osx_build: &geosx_osx_build
   - git lfs pull
   script:
   - GEOSX_TPL_DIR=/usr/local/GEOSX/GEOSX_TPL && sudo mkdir -p -m a=rwx ${GEOSX_TPL_DIR}/..
-  - python scripts/config-build.py -hc ${TRAVIS_BUILD_DIR}/host-configs/darwin-clang.cmake -bt Release -DNUM_PROC=$(nproc) -DGEOSXTPL_ENABLE_DOXYGEN:BOOL=OFF -ip ${GEOSX_TPL_DIR}
+  - python scripts/config-build.py -hc ${TRAVIS_BUILD_DIR}/host-configs/darwin-clang.cmake -bt Release -ip ${GEOSX_TPL_DIR} -DNUM_PROC=$(nproc) -DGEOSXTPL_ENABLE_DOXYGEN:BOOL=OFF -DENABLE_VTK:BOOL=OFF
```
rrsettgast, line 27: Did the order matter?

rrsettgast (approved):

.travis.yml
klevzoff: No, I just chose to group all `-xx` and all `-D` options

TotoGaz: You have no idea how dangerous this is https://github.com/GEOSX/GEOSX/pull/1832 🤣 😆

PR #187, opened 2022-05-11
Feature/total pangea3 env

XL64:
Adapt TPL to Pangea3 build and activate options for pygeosx (shared hdf5 + VTK wrappers)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,7 @@ option( ENABLE_TRILINOS "" ON )
 option( ENABLE_PETSC "" OFF )
 option( ENABLE_SUITESPARSE "" ON )
 option( ENABLE_VTK "" ON )
+option( ENABLE_VTK_WRAP_PYTHON "" OFF )
 
 option( ENABLE_NINJA "" OFF )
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -198,7 +199,7 @@ ExternalProject_Add( hdf5
                                        --enable-build-mode=production
                                        --prefix=<INSTALL_DIR>
                                        ${HDF5_ENABLE_PARALLEL}
-                                       --enable-shared=no
+                                       --enable-shared=yes
```
TotoGaz, line 202: Why do you need this?
Does it build both `static` and `shared objects`?

XL64: We need shared libs for pygeosx. I'll have to check.

XL64: It builds both .a and .so:
ls /workrd/SCR/NUM/shared-builds/p3/TPL/install/origin/master_187/install-pangea3-gcc8.4.1-openmpi-4.1.2-relwithdebinfo/hdf5/lib/libhdf5
libhdf5.a              libhdf5.la             libhdf5.settings       libhdf5.so             libhdf5.so.200         libhdf5.so.200.1.0     libhdf5_hl.a           libhdf5_hl.la          libhdf5_hl.so          libhdf5_hl.so.200      libhdf5_hl.so.200.0.1

sframba: @XL could we make this flag (and maybe also `ENABLE_VTK_WRAP_PYTHON`) conditional on the `ENABLE_PYGEOSX` cmake flag, so that we don't impact current non-pygeosx installs?

TotoGaz:
@rrsettgast @bmhan12 This PR looks good to me. Any additional check on LLNL side?

acitrain (approved):

PR #195, opened 2022-08-11
[WIP] Testing Github Actions

bmhan12:
This PR is testing out github actions to build Docker images.


bmhan12:
@rrsettgast @TotoGaz  Testing out Github Actions, the `docker build` portion looks to be working fine mostly. Github actions needs [security keys](https://docs.github.com/en/actions/security-guides/encrypted-secrets) to test out the [OSX build](https://github.com/GEOSX/thirdPartyLibs/blob/708482b6f870cabc884216629135e7b92f080236/.travis.yml#L25) and `docker push` to[ update DockerHub](https://github.com/GEOSX/thirdPartyLibs/blob/708482b6f870cabc884216629135e7b92f080236/.travis.yml#L55-L56).

scripts/docker_build_and_push.sh
```diff
@@ -0,0 +1,60 @@
...
+  # the `Formula/open-mpi.rb` history.)
+  BREW_HASH=fd64066273528a594e1f83c9eba20556e925e51c
+  BREW_URL=https://raw.github.com/Homebrew/homebrew-core/${BREW_HASH}
+  brew update > /dev/null 2>&1
+  wget ${BREW_URL}/Formula/open-mpi.rb
+  HOMEBREW_NO_AUTO_UPDATE=1 brew install ./open-mpi.rb
+  for dep in `brew deps open-mpi.rb`; do brew install $dep; done
+  wget ${BREW_URL}/Formula/git-lfs.rb
+  HOMEBREW_NO_AUTO_UPDATE=1 brew install ./git-lfs.rb
+  for dep in `brew deps git-lfs.rb`; do brew install $dep; done
```
TotoGaz, line 44: Note that I've just modified the this `brew` part, you should change it 👼

.github/workflows/docker_build_tpls.yml
```diff
@@ -0,0 +1,134 @@
...
+        echo "DOCKER_REPOSITORY=$docker_repository" >> $GITHUB_ENV
+        echo "DOCKER_COMPILER_BUILD_ARG=$docker_compiler_build_arg" >> $GITHUB_ENV
+        echo "OS=$os" >> $GITHUB_ENV
+        echo "INSTALL_DIR=$install_dir" >> $GITHUB_ENV
+        echo "HOST_CONFIG=$host_config" >> $GITHUB_ENV
+        echo "DOCKER_ROOT_IMAGE=$docker_root_image" >> $GITHUB_ENV
+        echo "PULL_REQUEST_NUMBER=$pull_request_number" >> $GITHUB_ENV
+        echo "BUILD_NUMBER=$build_number" >> $GITHUB_ENV
+        echo "COMMIT=$commit" >> $GITHUB_ENV
+        echo "BUILD_DIR=$build_dir" >> $GITHUB_ENV
```
TotoGaz, line 111: What is this? Is this for debugging or is it mandatory?

.github/workflows/docker_build_tpls.yml
```diff
@@ -0,0 +1,134 @@
...
+      fail-fast : false
+
+      matrix:
+        include:
+          - name: Pangea 2 (centos 7.6, gcc 8.3.0, open-mpi 2.1.5, mkl 2019.3)
+            DOCKER_REPOSITORY: han12/pangea2-gcc8.3.0-openmpi2.1.5-mkl2019.3
+            DOCKERFILE: docker/TotalEnergies/Dockerfile
+            DOCKER_ROOT_IMAGE: totogaz/pangea2-gcc8.3.0-openmpi2.1.5-mkl2019.3-no-geosx:0.0.1
+            INSTALL_DIR: /workrd/SCR/GEOSX/install/gcc8/GEOSX_TPL-${{ github.event.number }}-${{github.run_number}}-${{ github.sha }}
+            os: ubuntu-latest
```
TotoGaz, line 27: I think we should control our versioning be defining the exact version of our machines.

scripts/docker_build_and_push.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+env
+
+if [ "$OS" == "ubuntu-latest" ]
```
TotoGaz, line 4: Here and there, then maybe change with some `startswith` check?
We could even do some `if ( mac ) / else` instead of having a double `if`?

TotoGaz (commented):
What is our policy?
We'll end up with travis, github actions and azure.
Why do we need so many cloud tools? Looks like we'll have to spend time learning them?

scripts/docker_build_and_push.sh
bmhan12: Thanks, good catch. I need to update with `develop`.

.github/workflows/docker_build_tpls.yml
bmhan12: > What is this? Is this for debugging or is it mandatory?

This is one of those weird quirks about Github Actions.
Unlike Azure and Travis, environment variables you set are limit to the _job step_, not the _job_. So if I want to use an environment variable I've defined across all the steps in the job, I need to write it to `$GITHUB_ENV`. I also couldn't access Github Actions' defined environment variables in a bash script.

PR #209, opened 2022-11-11
update hypre and add cuda 11 builds

rrsettgast:
updated hypre
added Cuda 11 builds on LC

related to:
https://github.com/GEOSX/GEOSX/pull/2164

### DIFF
--- a/scripts/setupLC-TPL-helper.bash
+++ b/scripts/setupLC-TPL-helper.bash
@@ -25,8 +25,6 @@ echo "Building the TPLs for $HOST_CONFIG to be installed at $INSTALL_DIR. Progre
 ssh $MACHINE -t "
 . /etc/profile  &&
 cd $PWD &&
-export HYPRE_CUDA_SM=70 &&
-export CUDA_HOME=/usr/tce/packages/cuda/cuda-10.1.243 &&
 module load cmake/3.22.4 &&
 python scripts/config-build.py -hc $HOST_CONFIG -bt Release -ip $INSTALL_DIR $@ &&
 cd build-$CONFIG-release &&
--- a/scripts/setupLC-TPL-helper.bash
+++ b/scripts/setupLC-TPL-helper.bash
@@ -27,6 +27,7 @@ ssh $MACHINE -t "
 cd $PWD &&
 export HYPRE_CUDA_SM=70 &&
 export CUDA_HOME=/usr/tce/packages/cuda/cuda-10.1.243 &&
+module load cmake/3.22.4 &&
 python scripts/config-build.py -hc $HOST_CONFIG -bt Release -ip $INSTALL_DIR $@ &&
 cd build-$CONFIG-release &&
 $GET_A_NODE make &&
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -317,7 +317,7 @@ ExternalProject_Add( raja
                      PREFIX ${PROJECT_BINARY_DIR}/raja
                      INSTALL_DIR ${RAJA_DIR}
                      DEPENDS ${RAJA_DEPENDS}
-                     BUILD_COMMAND ${TPL_BUILD_COMMAND}
+                     BUILD_COMMAND ${TPL_BUILD_COMMAND} VERBOSE=1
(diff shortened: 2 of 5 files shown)


scripts/setupLC-TPL-helper.bash
```diff
@@ -27,6 +27,7 @@ ssh $MACHINE -t "
 cd $PWD &&
 export HYPRE_CUDA_SM=70 &&
 export CUDA_HOME=/usr/tce/packages/cuda/cuda-10.1.243 &&
```
victorapm, line 29: Do we want to set cuda-10 here? I see there are two new builds using cuda-11 in this PR.

victorapm (approved):
Thanks, Randy! I just had a minor question about setting `CUDA_HOME`

scripts/setupLC-TPL-helper.bash
rrsettgast: nice catch. if I recall, this was here because `hypre` needed it. is this still the case?

victorapm: That makes sense. Yes, hypre uses `CUDA_HOME` to find the paths for `nvcc` and the nvidia libraries. How about setting `CUDA_HOME` directly in the host-config `.cmake` file?

rrsettgast: Can we specify through cmake, or does hypre always look to the `CUDA_HOME` environment variable?

klevzoff: We are already providing `--with-cuda-home` and `--with-gpu-arch` to hypre's `configure` command in our `CMakeLists.txt`. I believe on Stanford/Sherlock we are building with both CUDA 10.1 and CUDA 11.2 without changing env variables.

victorapm: Just realized you are already using `--with-cuda-home`, that's another option. So, setting `CUDA_HOME` here is not necessary for hypre and can be deleted if not needed by other packages...

scripts/setupLC-TPL-helper.bash
```diff
@@ -27,6 +27,7 @@ ssh $MACHINE -t "
 cd $PWD &&
 export HYPRE_CUDA_SM=70 &&
```
victorapm, line 28:
```suggestion
```

This can be removed since it's being set here:
https://github.com/GEOSX/thirdPartyLibs/blob/ec70c19e82d0efe45b6d39d7df2ebca4fd36e975/CMakeLists.txt#L772

rrsettgast: It is still unclear to me if hypre still uses any environment variables at compile time and/or runtime?

victorapm: Hypre doesn't use environment variables at runtime.

During compilation of GPU builds, Hypre searches for the compute architecture in the environment variable `HYPRE_CUDA_SM`. However, if such information is passed explicitly to the configure script via the option `--with-gpu-arch`, hypre does not look for that environment variable. Note that geosx is currently using such an option: https://github.com/GEOSX/thirdPartyLibs/blob/ec70c19e82d0efe45b6d39d7df2ebca4fd36e975/CMakeLists.txt#L776

and `HYPRE_CUDA_SM` is defined here:
https://github.com/GEOSX/thirdPartyLibs/blob/ec70c19e82d0efe45b6d39d7df2ebca4fd36e975/CMakeLists.txt#L772

Thus, it is not necessary to define the environment variable `HYPRE_CUDA_SM` at the TPL bash script.

The same reasoning is valid for `CUDA_HOME`, since hypre is being configured with the option `--with-cuda-home`, see

https://github.com/GEOSX/thirdPartyLibs/blob/ec70c19e82d0efe45b6d39d7df2ebca4fd36e975/CMakeLists.txt#L773

It is safe to remove the definition of `HYPRE_CUDA_SM` and `CUDA_HOME` from the TPL bash script.

rrsettgast: ok. great.

bmhan12 (approved):
These changes look good to me, assuming the main repo runs on the new Docker images.

TotoGaz:
@XL64 can you trigger the build on Pangea3 ?

PR #212, opened 2022-12-13
Upgrade hypre version

TotoGaz:


docker/clang-cuda/Dockerfile
```diff
@@ -1,13 +1,16 @@
 ARG TMP_DIR=/tmp
-FROM nvidia/cuda:10.1-devel-ubuntu18.04 AS tpl_toolchain_intersect_geosx_toolchain
+FROM nvidia/cuda:11.2.2-devel-ubuntu20.04 AS tpl_toolchain_intersect_geosx_toolchain
+
+ENV TZ=America/Los_Angeles
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
```
TotoGaz, line 5:
```suggestion
RUN ln -fs /usr/share/zoneinfo/America/Los_Angeles /etc/localtime && \
    dpkg-reconfigure -f noninteractive tzdata
```
If it works (it actually works on my images), I'd do this instead, to reduce the footprint (you're actually adding an `ENV` which we may not want).

TotoGaz (commented):
I'd change the `tzdata` line.
Do not forget to update the `GEOSX_TPL_TAG` in the GEOSX PR accordingly.

francoishamon (approved):

PR #219, opened 2023-05-04
add `_d` suffix to `minizip` static library

tbeltzun:
When building the thirdPartyLibs using the CMake `Debug` configuration, the `minizip` target has the [DEBUG_POSTFIX](https://cmake.org/cmake/help/latest/prop_tgt/CONFIG_POSTFIX.html#prop_tgt:%3CCONFIG%3E_POSTFIX) property set to `_d`.

As a result, `fesapi` fails to find the correct library in https://github.com/GEOS-DEV/thirdPartyLibs/blob/92075d20def5608ae5777d8602b4555d2f384f63/CMakeLists.txt#L1248

This PR propagates the`minizip` postfix, allowing to build thirdPartyLibs in `Debug` configuration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1232,6 +1232,10 @@ if( ENABLE_FESAPI )
     set(FESAPI_URL "${TPL_MIRROR_DIR}/fesapi-2.4.0.0.tar.gz")
     message(STATUS "Building Fesapi found at ${FESAPI_URL}")
 
+    if( CMAKE_BUILD_TYPE MATCHES Debug )
+        set(minizip_postfix "_d")
+    endif()
+
     ExternalProject_Add( fesapi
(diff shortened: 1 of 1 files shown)


klevzoff:
This is surprising, given that we don't explicitly propagate `CMAKE_BUILD_TYPE` to `minizip` sub-project, nor does `ExternalProject_Add` inherit it implicitly (I know this because I've had to manually add `-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}` to multiple TPLs to get debug builds for them). By any chance, have you been able to figure out why the `minizip` sub-project gets built in debug in the first place?

untereiner:
The env variable check seems to be a recent  behavior. I wasn’t aware of it. To me there should be a findMinizip that handles the path

PR #221, opened 2023-05-10
Update fmt version

klevzoff:
Update `{fmt}` to 9.1.0

Related to https://github.com/GEOS-DEV/GEOS/pull/2426#issuecomment-1540981163


klevzoff:
Not necessary

Issue #43, opened 2019-06-27
Very hard to compile with "low" memory computer


AntoineMazuyer:
My memory get often saturated when compiling the tpls on my laptop (and I have 8GB of memory)

when compiling trilinos (yes, him again), here is the result of the `top` command
```
 PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
 4386 amazuyer  20   0  596360 337764  14184 R   6.3   4.2   0:07.19 cc1plus                             
 4352 amazuyer  20   0  594656 335092  14216 R   6.0   4.1   0:07.30 cc1plus                             
 4370 amazuyer  20   0  596612 338664  14216 R   6.0   4.2   0:07.23 cc1plus                             
 4373 amazuyer  20   0  588100 329648  14192 R   6.0   4.1   0:07.22 cc1plus                             
 4374 amazuyer  20   0  593540 332900  14140 R   6.0   4.1   0:07.22 cc1plus                             
 4384 amazuyer  20   0  593640 333604  14216 R   6.0   4.1   0:07.19 cc1plus                             
 4388 amazuyer  20   0  593712 334332  14196 R   6.0   4.1   0:07.18 cc1plus                             
 4391 amazuyer  20   0  591500 332324  14284 R   6.0   4.1   0:07.17 cc1plus                             
 4397 amazuyer  20   0  593608 334464  14140 R   6.0   4.1   0:07.17 cc1plus                             
 4399 amazuyer  20   0  596472 337268  14196 R   6.0   4.2   0:07.17 cc1plus                             
 4347 amazuyer  20   0  593680 336108  14140 R   5.6   4.1   0:07.30 cc1plus                             
 4371 amazuyer  20   0  596464 337088  14204 R   5.6   4.2   0:07.22 cc1plus                             
 4378 amazuyer  20   0  593632 332516  14220 R   5.6   4.1   0:07.20 cc1plus                             
 4393 amazuyer  20   0  591460 333656  14196 R   5.6   4.1   0:07.17 cc1plus                             
 4394 amazuyer  20   0  598564 340580  14216 R   5.6   4.2   0:07.17 cc1plus                             
 4398 amazuyer  20   0  596500 337308  14112 R   5.6   4.2   0:07.16 cc1plus
```

[...]

PR #222, opened 2023-05-12
Update raja chai vtk

rrsettgast:


TotoGaz:
@XL64 Can you trigger our jenkins on this PR?
BTW, we'll upgrade our cmake requirements to 3.23+. Is it OK on your side?

CMakeLists.txt
```diff
@@ -1071,7 +1073,7 @@ endif()
 ################################
 if( ENABLE_VTK )
     set( VTK_DIR "${CMAKE_INSTALL_PREFIX}/vtk" )
-    set( VTK_URL "${TPL_MIRROR_DIR}/VTK-9.1.0.tar.gz" )
+    set( VTK_URL "${TPL_MIRROR_DIR}/VTK-9.2.6.tar.gz" )
```
TotoGaz, line 1076: @klevzoff I do remember that you implemented some `vtk`-related code in GEOS specially because it was not public in `vtk` 9.1. Am I correct. Do you think some of this code could be cleaned?

klevzoff: Yes, it's related to cleaning up duplicate points:
https://github.com/GEOS-DEV/GEOS/blob/68c2ac12a80c55445d8c9ab260803245ecee0442/src/coreComponents/mesh/generators/VTKUtilities.cpp#L459-L461
and
https://github.com/GEOS-DEV/GEOS/blob/68c2ac12a80c55445d8c9ab260803245ecee0442/src/coreComponents/mesh/generators/VTKUtilities.cpp#L1809-L1815

It might take some manual labor to check whether the cleaning filter actually works the way I assumed it does. It can be done as a follow-up to the VTK update I think.

CMakeLists.txt
```diff
@@ -55,7 +55,7 @@ option( ENABLE_FESAPI "" ON )
 
 option( ENABLE_NINJA "" OFF )
 
-set(NUM_PROC "16" CACHE STRING "")
+set(NUM_PROC "36" CACHE STRING "")
```
klevzoff, line 58: This is a bit tough on unsuspecting laptop users who don't explicitly set `NUM_PROC` :laughing: That said, neither was the old default good for them, so...

CMakeLists.txt
```diff
@@ -322,14 +322,13 @@ ExternalProject_Add( raja
...
-                                -D CMAKE_CXX_FLAGS:PATH=${CMAKE_CXX_FLAGS}
                                 -D CMAKE_BUILD_TYPE:STRING=${CMAKE_BUILD_TYPE}
-                                -D CMAKE_CXX_FLAGS_RELEASE:STRING=${CMAKE_CXX_FLAGS_RELEASE}
                                 -D CMAKE_VERBOSE_MAKEFILE:BOOL=${CMAKE_VERBOSE_MAKEFILE}
                                 -D ENABLE_CUDA:BOOL=${ENABLE_CUDA}
+                                -D RAJA_ENABLE_CUDA:BOOL=${ENABLE_CUDA}
                                 -D CMAKE_CUDA_COMPILER=${CMAKE_CUDA_COMPILER}
                                 -D CUDA_TOOLKIT_ROOT_DIR=${CUDA_TOOLKIT_ROOT_DIR}
-                                -D CUDA_ARCH:CACHE=${CUDA_ARCH}
+                                -D CMAKE_CUDA_ARCHITECTURES:STRING=${CMAKE_CUDA_ARCHITECTURES}
```
klevzoff, line 331: Consider setting
```cmake
cmake_minimum_required(VERSION 3.18)
```
at the top of the file, or this will get ignored. We have 3.16 and 3.17 users for sure (3.16 is default on ubuntu-20.04). They will be forced to upgrade, but it's better than silently building for the wrong arch.

tplMirror/vtkXMLReader-fpe.patch
```diff
@@ -0,0 +1,11 @@
+--- IO/XML/vtkXMLReader.cxx	2023-02-14 20:03:53.000000000 -0800
++++ IO/XML/vtkXMLReader.cxx.patch	2023-05-12 08:52:36.017274000 -0700
+@@ -1958,7 +1958,7 @@
+   if (!this->AbortExecute)
+   {
+     // Round progress to nearest 100th.
+-    float rounded = static_cast<float>(int((progress * 100) + 0.5f)) / 100.f;
++    float const rounded = static_cast<float>(static_cast<int>((progress * 100) + 0.5)) * 0.01;
```
klevzoff, line 8: Nice one! Do we want to try upstream this fix?

(Now I'm also curious what value of `progress` caused this to... underflow?)

klevzoff (approved):

CMakeLists.txt
klevzoff: Seems like RAJA is already requiring 3.20 minimum?

tplMirror/vtkXMLReader-fpe.patch
rrsettgast: `0`
I actually don't understand why this was an FPE. Do you understand what was happening?

CMakeLists.txt
rrsettgast: oh...I meant to set 3.23.1 as the minimum.

tplMirror/vtkXMLReader-fpe.patch
klevzoff: Apparently, truncating any floating point value with non-zero fractional part to `int` causes `FE_INEXACT` to be raised: https://godbolt.org/z/sWrjf5vns

I don't really understand how this worked before and still works on other platforms. In the above godbolt link the FPE can be reproduced under any compiler and any optimization level. It's also not clear to me why your patch fixes it? The proper way to round down should be `std::floor`. Floating point is still mystery to me sometimes...

CMakeLists.txt
```diff
@@ -55,7 +55,12 @@ option( ENABLE_FESAPI "" ON )
 
 option( ENABLE_NINJA "" OFF )
 
-set(NUM_PROC "16" CACHE STRING "")
+include(ProcessorCount)
+ProcessorCount(NUM_CORES)
+			    
+set(NUM_PROC ${NUM_CORES} CACHE STRING "")
+message( "Running with NUM_PROC = ${NUM_PROC}")
```
rrsettgast, line 62: @klevzoff Did we ever try using cmake to set `NUM_PROC` like this?

klevzoff: I don't think so, though you've suggested this before :) https://github.com/GEOS-DEV/thirdPartyLibs/issues/43#issuecomment-525394012

tplMirror/vtkXMLReader-fpe.patch
rrsettgast: Oh. I think `std::round(progress*100)*0.01` is what they are trying to do here. I was wondering why they don't just use one of the available functions for this operation. Also, they then take this float and to a floating point `!=` in the next line. Don't know why things are done this way.

CMakeLists.txt
rrsettgast: how embarrassing that I was that lazy even before COVID.

untereiner:
I have a compilation with gcc 13 and clang15 [issue](https://gitlab.kitware.com/vtk/vtk/-/commit/b1a09529f3ab381494f3eecb53b2ce4adc5b55e9#). A fix will be [merged](https://gitlab.kitware.com/vtk/vtk/-/merge_requests/10268#note_1375021) in vtk 9.3.0 that should be released on june 15th.

I have also issues with chai

PR #223, opened 2023-05-30
add patch file. Add a couple of cmake CUDA changes

rrsettgast:
Add a patch to avoid FPE when in vtk output routines.

also added a couple of changes to the flags passed to raja.


CusiniM (approved):

klevzoff:
What about https://github.com/GEOS-DEV/thirdPartyLibs/pull/222 ? It's not getting merged?

rrsettgast:
> What about #222 ? It's not getting merged?

Not at the moment. Having some pretty annoying problems on LC with the updated packages. I think #222 will go in a more incremental fashion. Also ditching the cuda10 compiler on lassen will avoid some problems, so perhaps after the c++17 update it will just work.

PR #225, opened 2023-06-22
Github Actions Pipeline

bmhan12:
This PR:
- Creates a Github Actions version of the Travis CI workflow

TODO:
- [x] Test deployment to Dockerhub
    - [x] ~~Update macosx_TPL_mngt.py script~~
    - [x] Remove mac OSX build
- [x] Test Docker containers with GEOS main repo CI pipeline

Related PR: GEOS-DEV/GEOS#2573


.github/workflows/docker_build_tpls.yml
```diff
@@ -0,0 +1,114 @@
...
+      # In-progress jobs will not be cancelled if there is a failure
+      fail-fast : false
+
+      matrix:
+        include:
+          - name: Pangea 2 (centos 7.6, gcc 8.3.0, open-mpi 2.1.5, mkl 2019.3)
+            DOCKER_REPOSITORY: geosx/pangea2-gcc8.3.0-openmpi2.1.5-mkl2019.3
+            DOCKERFILE: docker/TotalEnergies/Dockerfile
+            DOCKER_ROOT_IMAGE: totogaz/pangea2-gcc8.3.0-openmpi2.1.5-mkl2019.3-no-geosx:0.0.1
+            INSTALL_DIR: /workrd/SCR/GEOSX/install/gcc8/GEOSX_TPL-${{ github.event.number }}-${{github.run_number}}-${{ github.sha }}
```
TotoGaz, line 27: In the installation dirs, is it possible to replace the `${{ github.sha }}` into `${GITHUB_SHA:0:7}`?
Very long folders are more complicated to manage.

I could not find a way to use the `github` context simply (actually, `sha_short` was not out of the box 🤣 ).

TotoGaz (commented):
I think that you can now remove the `.travis.yml` file 😨 😱 !!!!

scripts/docker_build_and_push.sh
```diff
@@ -0,0 +1,30 @@
...
+--build-arg INSTALL_DIR=${INSTALL_DIR} \
+--tag ${DOCKER_REPOSITORY}:${DOCKER_TAG} \
+--file ${DOCKERFILE} \
+--label "org.opencontainers.image.created=$(date --rfc-3339=seconds)" \
+--label "org.opencontainers.image.source=https://github.com/GEOSX/thirdPartyLibs" \
+--label "org.opencontainers.image.revision=${COMMIT}" \
+--label "org.opencontainers.image.title=Building environment for GEOSX" \
+.
+
+docker push ${DOCKER_REPOSITORY}:${DOCKER_TAG}
```
TotoGaz, line 30: You're also pushing the image, for the same PR, but with a build id different (one is a travis build, the other is a GHA build), in the same `docker` repository.
Isn't it a bit dangerous? Is there a reason why you're not simply removing `travis`? What is the objective of this PR?

.github/workflows/docker_build_tpls.yml
```diff
@@ -0,0 +1,114 @@
...
+      # In-progress jobs will not be cancelled if there is a failure
+      fail-fast : false
+
+      matrix:
+        include:
+          - name: Pangea 2 (centos 7.6, gcc 8.3.0, open-mpi 2.1.5, mkl 2019.3)
+            DOCKER_REPOSITORY: geosx/pangea2-gcc8.3.0-openmpi2.1.5-mkl2019.3
+            DOCKERFILE: docker/TotalEnergies/Dockerfile
+            DOCKER_ROOT_IMAGE: totogaz/pangea2-gcc8.3.0-openmpi2.1.5-mkl2019.3-no-geosx:0.0.1
+            INSTALL_DIR_ROOT: /workrd/SCR/GEOSX/install/gcc8/GEOSX_TPL
```
TotoGaz, line 27: The original paths contain some information on the build (commit hash, PR...). Like `INSTALL_DIR=/workrd/SCR/GEOSX/install/gcc8/GEOSX_TPL-${TRAVIS_PULL_REQUEST}-${TRAVIS_BUILD_NUMBER}-${TRAVIS_COMMIT:0:7}`.
I believe we can use the `[GITHUB_SHA](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)` env variable to select the first 7 digits.
This is important not to have deployments conflicts (they're supposed to all live in their own folder).

scripts/docker_build_and_push.sh
```diff
@@ -0,0 +1,30 @@
...
+# This script will build and push a DOCKER_REPOSITORY:DOCKER_TAG image build from DOCKERFILE
+# with (optional) DOCKER_COMPILER_BUILD_ARG build arguments.
+# A specific host-config file can be defined through variable HOST_CONFIG.
+# For the case of Total cluster only, DOCKER_ROOT_IMAGE is used to define docker base image.
+# Where the TPL are installed in the docker can be specified by parameter INSTALL_DIR.
+# Unlike DOCKER_TAG, these variables shall be defined by the "yaml derived classes" in a stage prior to `script` stage.
+DOCKER_TAG=${PULL_REQUEST_NUMBER}-${BUILD_NUMBER}
+echo "Docker tag is ${DOCKER_REPOSITORY}:${DOCKER_TAG}"
+
+INSTALL_DIR=${INSTALL_DIR_ROOT:-/opt/GEOSX/GEOSX_TPL}-${PULL_REQUEST_NUMBER}-${BUILD_NUMBER}-${COMMIT:0:7}
```
bmhan12, line 15: > The original paths contain some information on the build (commit hash, PR...)

The path information is still preserved here. 
It's just been moved to the script.

scripts/docker_build_and_push.sh
bmhan12: > Isn't it a bit dangerous? Is there a reason why you're not simply removing travis? What is the objective of this PR?

Sorry, I wasn't sure if it was possible to do all of the necessary steps in one PR:

- Pass the current required checks for PR's
- Disable Travis CI as a "required" check
- Remove the Travis CI workflow file
- Add the new Github Actions check in this PR as a "required" check for PR's
- Add the Github Actions workflow file

In my experience, it shows Travis CI as permanently "pending" if I remove the `travis.yml` , as it's still "required" CI for PR's. So it is still necessary to have to pass the PR checks.

As an aside, I don't think the build id's can conflict, as with Docker Tag `XX:YY`, the `XX` value is the same for both Github Actions and Travis CI, the PR number. But, the second value for Github Actions is the current quantity of commits in the PR (e.g. 44 for this current PR), but for Travis CI it is a rolling sum of all builds made by Travis CI (in GEOS, this value is 965).

scripts/docker_build_and_push.sh
TotoGaz: Oh, I did not catch it sorry!
What do you think of changing
```
INSTALL_DIR=${INSTALL_DIR_ROOT:-/opt/GEOSX/GEOSX_TPL}-${PULL_REQUEST_NUMBER}-${BUILD_NUMBER}-${COMMIT:0:7}
```
into
```
INSTALL_DIR=${INSTALL_DIR_ROOT:-/opt/GEOSX}/GEOSX_TPL-${PULL_REQUEST_NUMBER}-${BUILD_NUMBER}-${COMMIT:0:7}
```
or something like this?
Then
```
INSTALL_DIR_ROOT: /data/gpfs/Users/j0436735/travis-deployments/GPU/GEOSX_TPL
```
can be changed into
```
INSTALL_DIR_ROOT: /data/gpfs/Users/j0436735/travis-deployments/GPU
```
which is more "classic" in the sense where `INSTALL_DIR_ROOT` is really a folder?
The result should be identical?

scripts/docker_build_and_push.sh
TotoGaz: > As an aside, I don't think the build id's can conflict, as with Docker Tag XX:YY, the XX value is the same for both Github Actions and Travis CI, the PR number. But, the second value for Github Actions is the current quantity of commits in the PR (e.g. 44 for this current PR), but for Travis CI it is a rolling sum of all builds made by Travis CI across all runs (in GEOSX_TPL_TAG, this value is 965).

Yes, I was a bit afraid about this numbering issue, but from what you say, as long as we do not have 965 commits it should be OK. We will have it solved since then!

TotoGaz (approved):
I'm fine getting this merged. I do not thing we should dwell on having both travis and GHA transition, but it's very close now!

scripts/docker_build_and_push.sh
TotoGaz: Maybe we should even impose having
```
INSTALL_DIR_ROOT: /opt/GEOSX
```
in the `yml` file, for all the cases, so we get rid of the default parameter management? What do you think?

bmhan12: > Maybe we should even impose having...

I think that's good for consistency, eliminates need to go to the github actions configuration file to figure out where the installation is located.

scripts/docker_build_and_push.sh
```diff
@@ -12,7 +12,7 @@ echo .git > .dockerignore
 DOCKER_TAG=${PULL_REQUEST_NUMBER}-${BUILD_NUMBER}
 echo "Docker tag is ${DOCKER_REPOSITORY}:${DOCKER_TAG}"
 
-INSTALL_DIR=${INSTALL_DIR_ROOT:-/opt/GEOSX/GEOSX_TPL}-${PULL_REQUEST_NUMBER}-${BUILD_NUMBER}-${COMMIT:0:7}
+INSTALL_DIR=/opt/GEOSX/GEOSX_TPL-${PULL_REQUEST_NUMBER}-${BUILD_NUMBER}-${COMMIT:0:7}
```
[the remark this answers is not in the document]
bmhan12: @TotoGaz Woops! Think I got what you meant this time 😂 .

scripts/docker_build_and_push.sh
```diff
@@ -0,0 +1,30 @@
...
+# This script will build and push a DOCKER_REPOSITORY:DOCKER_TAG image build from DOCKERFILE
+# with (optional) DOCKER_COMPILER_BUILD_ARG build arguments.
+# A specific host-config file can be defined through variable HOST_CONFIG.
+# For the case of Total cluster only, DOCKER_ROOT_IMAGE is used to define docker base image.
+# Where the TPL are installed in the docker can be specified by parameter INSTALL_DIR.
+# Unlike DOCKER_TAG, these variables shall be defined by the "yaml derived classes" in a stage prior to `script` stage.
+DOCKER_TAG=${PULL_REQUEST_NUMBER}-${BUILD_NUMBER}
+echo "Docker tag is ${DOCKER_REPOSITORY}:${DOCKER_TAG}"
+
+INSTALL_DIR==${INSTALL_DIR_ROOT}/GEOSX_TPL-${PULL_REQUEST_NUMBER}-${BUILD_NUMBER}-${COMMIT:0:7}
```
TotoGaz, line 15: Are the double `=` signs intended?

TotoGaz (approved):

scripts/docker_build_and_push.sh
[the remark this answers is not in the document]
TotoGaz: 😉

scripts/docker_build_and_push.sh
bmhan12: > Are the double = signs intended?

Definitely not! Good catch.

rrsettgast (approved):

PR #227, opened 2023-07-10
fix boolean comparison

tbeltzun:
`CMake` [supports additional constants](https://cmake.org/cmake/help/latest/command/if.html#basic-expressions) other than `ON / OFF` in if clauses.

Using a hardcoded constant is a mistake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ cmake_minimum_required(VERSION 3.23.1)
 project(geosx_tpl LANGUAGES C CXX Fortran)
 
 option( GEOSXTPL_ENABLE_DOXYGEN "" ON )
-if ( ENABLE_DOXYGEN EQUAL OFF )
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ cmake_minimum_required(VERSION 3.23.1)
 project(geosx_tpl LANGUAGES C CXX Fortran)
 
 option( GEOSXTPL_ENABLE_DOXYGEN "" ON )
-if ( ENABLE_DOXYGEN EQUAL OFF )
+if ( ENABLE_DOXYGEN )
     set( GEOSXTPL_ENABLE_DOXYGEN "OFF" )
 endif()
```
TotoGaz, line 10: Are you sure that we really want to do `if enable doxygen then set ENABLE_DOXYGEN to off`?

CusiniM (approved):
Thanks for catching this! It was kind of driving me crazy

PR #228, opened 2023-07-12
fix `CMAKE_VERBOSE_MAKEFILE`

tbeltzun:
Fix bug introduced by https://github.com/GEOS-DEV/thirdPartyLibs/pull/55.

The[ `set` syntax](https://cmake.org/cmake/help/latest/command/set.html#set-cache-entry) is inverted, leading to a non-verbose build (e.g. raja).


Algiane:
Hi,

By curiosity: why is it needed to set explicitely the `CMAKE_VERBOSE_MAKEFILE` variable to `OFF` in the CMake process (according to the [CMake documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_VERBOSE_MAKEFILE.html), this variable is automatically added to the cache variables and its default value is `OFF` as wanted here)?

Best

Algiane (approved):

PR #235, opened 2023-08-14
Remove TravisCI dependency

rrsettgast:


cmake/blt
TotoGaz, line 1: Is the update of the submodule intended?

rrsettgast: no

TotoGaz: So why don't we remove it?

bmhan12 (approved):

PR #239, opened 2023-09-15
Build hdf5 using CMake instead of Autogen.

### LINKED ISSUE #211: configure HDF5 with CMake intead of configure
Hi,
When using configure/make the cmake configuration files are not installed (HDF5config.cmake and HDF5options.cmake).
Thus, a cmakefile relying completelly on FindHDF5.cmake and HDF5_ROOT or HDF5_DIR will fail to find the TPL one.
It will possibly find a hdf5 already installed on the system because h5cc or h5pcc from the system is called.

Algiane:
It avoids the spurious detection of the NDEBUG compilation flag by CMake when importing the hdf5 package.

As a side-effect, it will allow to test the GEOS `asserts` in  github-ci configs running in `Debug` mode.

See:
  - Geos issue [#2631](https://github.com/GEOS-DEV/GEOS/issues/2631)
  - hdf5 issue [#3526](https://github.com/HDFGroup/hdf5/issues/3526#issuecomment-1719664682)
  

Closes #211, closes https://github.com/GEOS-DEV/GEOS/issues/2631, closes https://github.com/HDFGroup/hdf5/issues/3526.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -188,29 +188,33 @@ message(STATUS "Building HDF5 found at ${HDF5_URL}")
 if( ${ENABLE_MPI} )
   set( HDF5_C_COMPILER ${MPI_C_COMPILER} )
   set( HDF5_CXX_COMPILER ${MPI_CXX_COMPILER} )
-  set( HDF5_ENABLE_PARALLEL "--enable-parallel")
 else()
   set( HDF5_C_COMPILER ${CMAKE_C_COMPILER} )
   set( HDF5_CXX_COMPILER ${CMAKE_CXX_COMPILER} )
-  set( HDF5_ENABLE_PARALLEL "")
 endif()
-
-set(HDF5_C_FLAGS "-fPIC ${C_FLAGS_NO_WARNINGS} ${CMAKE_C_FLAGS_RELEASE}")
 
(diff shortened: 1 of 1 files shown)


untereiner:
Should also close #211

untereiner (commented):
just a comment, not sure if it matters but other external projects have 
`-D CMAKE_BUILD_TYPE:STRING=${CMAKE_BUILD_TYPE}`

Algiane:
Hi @untereiner ,

Thanks for your feedbacks: I have tried to preserve exactly the options that where used to build hdf5 using autogen. As it was always built in `production` mode, I have chosen to not transfer the `CMAKE_BUILD_TYPE` info to hdf5 and to always build it in `Release` cmake build type. 

In practice I don't really have an opinion on what is better: do we want to build all TPLs in Debug mode sometimes or only some of them?

Algiane:
Ok, So I will transfer the global CMAKE_BUILD_TYPE toward hdf5.

Just for your info: no that I have commit rights in the repo, I will redo this PR from a branch instead rather than a fork (it seems that the ci doesn't work if launched by a fork due to missing identifiers for connecting to DockerHub.

I will link this conversation to the new PR.

PR #240, opened 2023-09-15
Fix wrong setting of CMAKE_VERBOSE_MAKEFILE variable.

Algiane:
The CMAKE_VERBOSE_MAKEFILE variable was setted to `BOOL;OFF;` due to an erroneous call of the SET function of CMake (for non cache variable, the type of the variable is not expected).

It was leading to the following warning when compiling some of the TPLs (for example adiak with `VERBOSE=1 make` command):
```
CMake Warning:
  Ignoring extra path from command line:

   "/Users/froehly/GEOS-FORK/thirdPartyLibs-AlgianeFork/build-appleM1-clang-debug/adiak/src/adiak-build/OFF"
```


Algiane:
I close this PR as I have just seen that the PR#228 fixes the same issue.

PR #253, opened 2023-12-22
Updated Pangea3 host config with C++17 standard

sframba:
Companion PR to GEOS' PR [#2910](https://github.com/GEOS-DEV/GEOS/pull/2910), this PR updates the Pangea 3 host config file after upgrade od `cuda` to version `11.5` on this machine. In particular,
 - The c++ standard is set to c++17 (as it should be for GEOS)
 - ~~The default hypre implementation on the machine is set to CPU (GPU not supported for all applications)~~ hypre with cuda acceleration seems to be mostly working now
 - A flag allowing host constexpr call form device code is added, reducing warnings 
 - CUpti profiling via Caliper is enabled by default


victorapm:
Hi Stefano, can you share which models doesn't work with hypre GPU?

sframba:
> Hi Stefano, can you share which models don't work with hypre GPU?

Hi again @victorapm, it seems that with the latest version of hypre, one of the GEOS unit tests still fails: `testExternalSolvers`, with the follwoing test failing: `Hypre/SolverTestLaplace2D/0.GMRES_ILU`.
Still, the flow simulation that was holding us back seems to be working now, so I'll re-enable hypre cuda acceleration by default in this PR

victorapm:
Thanks, Stefano! I will check with @castelletto1 about the failing test

TotoGaz (approved):

PR #259, opened 2024-02-08
Pygeosx Testing

bmhan12:
This PR attempts to install pygeosx dependencies to run pygeosx tests.


docker/gcc-ubuntu/Dockerfile
```diff
@@ -120,7 +121,17 @@ RUN apt-get install -y --no-install-recommends \
...
+    python3 \
+    python3-dev \
+    python3-pip \
+    python3-numpy \
+    python3-mpi4py \
+    python3-scipy \
+    python3-virtualenv \
+    python3-matplotlib \
+    python3-venv \
+    python3-pytest
```
CusiniM, line 134: are these all needed by pegeosx?

CusiniM: doesn't scipy already install `numpy` and and `matplotlib` ?

CusiniM (approved):

docker/gcc-ubuntu/Dockerfile
bmhan12: > are these all needed by pygeosx?

Yes, as far as I know. These were based on a list of dependencies I received from @cssherman when trying to build pygeosx dependencies with spack.

> doesn't scipy already install numpy and and matplotlib ?

Good point. I went through the dependency list and removed any dependency of dependencies that would be installed. Turns out that [scipy includes numpy, but not matplotlib](https://packages.ubuntu.com/jammy/python3-scipy).

.github/workflows/docker_build_tpls.yml
```diff
@@ -78,16 +78,16 @@ jobs:
...
     - name: Checkout
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4.1.1
       with:
         submodules: true
         lfs: true
 
     - name: Set up Docker Buildx
-      uses: docker/setup-buildx-action@v2
+      uses: docker/setup-buildx-action@v3
```
TotoGaz, line 87: @untereiner do you think a `dependabot` would be useful here as well?
If so, do you mind setting one up?

docker/gcc-ubuntu/Dockerfile
```diff
@@ -120,7 +121,15 @@ RUN apt-get install -y --no-install-recommends \
     libxml2-utils \
     git \
     ghostscript \
-    ninja-build
+    ninja-build \
+    python3 \
```
TotoGaz, line 125: I'm not a huge fan of having `python` hopping in our builds, even for non `pygeosx` builds... I guess it is what it is.

TotoGaz (approved):

rrsettgast:
@cssherman Does this do what you need?

PR #260, opened 2024-03-06
fix `TPL` builds with `clang >= 15`

tbeltzun:
As discussed on slack, this PR:

-  allows building the TPLs with `clang >= 15`, tested on linux, with clang `17.0.6 `(with or without cuda `12`) or  gcc `12.3.0` (with or without cuda `12`);
- cleans the propagation of `CFLAGS` to silo by writing a configuration file instead;
- adds a CI `geosx/ubuntu22.04-clang15` docker image based on clang `15.0.7` (per https://github.com/GEOS-DEV/GEOS/pull/3024#issuecomment-1989539601);
- update `CMake` and `sccache` versions to the latest available, and use a common script instead of hardcoding them into `Dockerfile`;
- use `RUN --mount=src=.,dst=$TSRC_DIR $SRC_DIR/foo.sh` instead of the `COPY . $SRC_DIR` + `RUN $SRC_DIR/foo.sh` hack;
- fixes a parsing bug in `configure_tpl_build.sh`: `[[` is `bash` syntax, **not** `sh` syntax;
- update job version labels (compilers);
- reorganization of `Dockerfile`s;
- renames and cleanup.

Reference https://www.redhat.com/en/blog/new-warnings-and-errors-clang-16.

~~TODO: we could probably move installing dependencies in `Dockerfile` to a common shell script ...~~


CusiniM (approved):
Thank you for this. Also fixes issues with building on macOS. Don't forge to create a GEOS PR with an updated tpl tag.

docker/clang-ubuntu/Dockerfile
```diff
@@ -0,0 +1,114 @@
...
+    ca-certificates \
+    curl \
+    libtbb2 \
+    libblas-dev \
+    liblapack-dev \
+    zlib1g-dev \
+    openmpi-bin \
+    libopenmpi-dev \
+# Some of the TPL's make "extensive" use of python in their build.
+# And we want to test GEOSX's python configuration script.
```
CusiniM, line 45:
```suggestion
# And we want to test GEOS's python configuration script.
```

docker/clang-ubuntu/Dockerfile
```diff
@@ -0,0 +1,114 @@
...
+# Unfortunately argparse (standard library's package used by GEOSX)
```
CusiniM, line 46:
```suggestion
# Unfortunately argparse (standard library's package used by GEOS)
```

docker/clang-cuda/Dockerfile
```diff
@@ -1,14 +1,18 @@
-ARG TMP_DIR=/tmp
 FROM nvidia/cuda:11.8.0-devel-ubuntu20.04 AS tpl_toolchain_intersect_geosx_toolchain
 
 ARG INSTALL_DIR
-ENV GEOSX_TPL_DIR=${INSTALL_DIR}
+ENV GEOSX_TPL_DIR=$INSTALL_DIR
+
+ENV TMP_DIR=/tmp
+ENV TPL_SRC_DIR=$TMP_DIR/thirdPartyLibs
+ENV TPL_BUILD_DIR=$TMP_DIR/build
```
CusiniM, line 8: was there a reason behind the original choice of defining these as build variables?

CusiniM (approved):
Thanks for this work. It looks good to me but I think it would be good if @TotoGaz also had a look at it. To me it looks like this PR also does a nice clean up but I am not necessarily sure of the consequences (maybe be none) of some of the changes to the dockerfiles. 

Btw, whenever we modify these we should also slowly change all `GEOSX` to `GEOS`.

docker/install_sccache.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+set -xe
+
+# Install `sccache` binaries to speed up the build of `geos`
+PREFIX=/opt/sccache
+VERSION=0.7.7
```
TotoGaz, line 6: Be sure that `sccache` is really working for cuda builds. There was an issue with previous versions, so this is something we should not be missing!

TotoGaz (approved):

docker/gcc-ubuntu/Dockerfile
```diff
@@ -11,7 +12,12 @@ FROM ${DOCKER_ROOT_IMAGE} AS tpl_toolchain_intersect_geosx_toolchain
 # The installation directory is provided as a docker build argument.
 # We forward it using an environment variable.
 ARG INSTALL_DIR
-ENV GEOSX_TPL_DIR=${INSTALL_DIR}
+ENV GEOSX_TPL_DIR=$INSTALL_DIR
+
+# Temporary local variables dedicated to the TPL build
+ENV TMP_DIR=/tmp
+ENV TPL_SRC_DIR=$TMP_DIR/thirdPartyLibs
+ENV TPL_BUILD_DIR=$TMP_DIR/build
```
TotoGaz, line 20: The problem with setting these variables as `ENV` is that they will (I think) become part of the environment of the final "product". The environment you can use is described [here](https://geosx-geosx.readthedocs-hosted.com/en/latest/docs/sphinx/buildGuide/ContinuousIntegration.html#docker-images-contract).
Still, it's better if we can hide those.

This is also why they were described using `ARG` previously.
Do you think that defining them using `ARG` (and then, if you have to, "importing" them using `ARG VAR`) could be OK? (same comment for the new `clang-ubuntu/Dockerfile`)

docker/Stanford/Dockerfile
```diff
@@ -1,23 +1,27 @@
+# NOTE: see docker/tpl-ubuntu-gcc.Dockerfile for detailed comments
```
TotoGaz, line 1: Yes 😉

TotoGaz (approved):
Thank you @tbeltzun I think it's good to go. Thank you for your work.

Now that you have been through this part of the CI, do you have a REX (a few bullet points ) of what we could improve?
And maybe points that you think were nice as well?

TotoGaz:
Yes, we had some duplications when installing tools like `sccache`. It's good that you were able to fix this.
Now, the global pattern is still duplicated accros the `Dockerfiles`, and some package names as well.
In the very very very beginning we had a pattern with a lot of scripts, allowing not to duplicated "anything".
But in the end, `docker` being what it is (_i.e._ not real programming language), the situation was extremely complicated to handle and understand. I believe that the current situation is a not-so-bad compromise given the size of the project. 🤷

PR #261, opened 2024-03-07
[WIP] Spack docker containers

bmhan12:
This PR attempts to build dependencies using uberenv, spack.


rrsettgast (approved):

wrtobin (approved):

wrtobin (approved):

docker/tpl-centos-gcc-cuda.Dockerfile
```diff
@@ -72,41 +69,51 @@ RUN yum install -y \
...
-RUN --mount=src=.,dst=$SRC_DIR make
+# Run uberenv
+# Have to create install directory first for uberenv
+# -k flag is to ignore SSL errors
+RUN --mount=src=.,dst=$SRC_DIR,readwrite cd ${SRC_DIR} && \
+     mkdir -p ${GEOSX_TPL_DIR} && \
+# Create symlink to openmpi include directory
+     ln -s /usr/include/openmpi-x86_64 /usr/lib64/openmpi/include && \
+     ./scripts/uberenv/uberenv.py \
+       --spec "%gcc@8+cuda~uncrustify~openmp~pygeosx cuda_arch=70 ^cuda@11.8.0+allow-unsupported-compilers ^caliper~gotcha~sampler~libunwind~libdw~papi" \
```
CusiniM, line 85: what' the meaning of these ~? Is that the list of tpls we are actually building? couldn't this be that we build all of them by default? I might be missing something.

corbett5: The first part `%gcc@8+cuda~uncrustify~openmp~pygeosx` is the spec for GEOS. We are building using GCC 8 with cuda, but without uncrustify, openmp or pygeosx.

The last part`^caliper~gotcha~sampler~libunwind~libdw~papi` means that we are building Caliper with the gotcha, sampler, libunwind, libdw, and papi options turned off.

CusiniM (approved):
I was able to build on my mac using spack (https://github.com/GEOS-DEV/thirdPartyLibs/pull/291/files#). I had a couple of issue but I think that we could go ahead and merge this. I would also like to propose to move all the spack build related files and the spack configs outside of the script folder. 

I also wonder whether we want to keep the tpls in a separate repo or not. Once we fully move to spack it may not really be necessary.

PR #264, opened 2024-05-16
Build Geos against TPLs with ENABLE_CUDA=OFF on pangea3

Algiane:
Using the ThirdPartyLibraries with the `pangea3-gcc8.4.1-openmpi-4.1.2.cmake` host-config file and the `-DENABLE_CUDA=OFF` option compiles HYPRE with CUDA flags and the other TPLs without CUDA flags. This results in a set of software that can't be used to build Geos.

This PR proposes to ensure that the entire suite of TPLs is built with compatible flags (so the build of Geos will not fail): 
  1. if the `ENABLE_CUDA` variable is explicitely setted to `OFF` by the user and `Hypre` is asked to be built with `CUDA`, a warning is raised saying that Hypre will ignore the `CUDA` config.
  2. Even if useless with the previous modification, I find it useful to provide to the user host-config files where dependencies between options are explicitely shown so I have modified the P3 host-config file in order to set `Hypre device` to `CUDA` only when `ENABLE_CUDA` is `on`.

It is linked to:
  -  Geos issue : https://github.com/GEOS-DEV/GEOS/issues/3128
  - Geos PR : https://github.com/GEOS-DEV/GEOS/pull/3129

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -797,9 +797,12 @@ if( ENABLE_HYPRE )
   endif()
 
   if( ENABLE_HYPRE_DEVICE STREQUAL "CUDA" )
-    list( APPEND HYPRE_DEPENDS chai )
-    string( SUBSTRING ${CUDA_ARCH} 3 -1 HYPRE_CUDA_SM ) # sm_XY -> XY
-    set( HYPRE_CUDA_FLAGS "\
+    if( NOT ENABLE_CUDA )
+      message( WARNING "Setting of HYPRE device to CUDA ignored because ENABLE_CUDA is OFF" )
+    else()
+      list( APPEND HYPRE_DEPENDS chai )
+      string( SUBSTRING ${CUDA_ARCH} 3 -1 HYPRE_CUDA_SM ) # sm_XY -> XY
+      set( HYPRE_CUDA_FLAGS "\
 --with-cuda --with-cuda-home=${CUDA_TOOLKIT_ROOT_DIR} \
 --enable-cusparse \
 --enable-cusolver \
@@ -809,7 +812,7 @@ if( ENABLE_HYPRE )
 --with-umpire-include=${CHAI_DIR}/include \
 --with-umpire-lib-dirs=${CHAI_DIR}/lib \
 --with-umpire-libs=umpire " )
-
+    endif()
   elseif( ENABLE_HYPRE_DEVICE STREQUAL "HIP" )
     message( FATAL_ERROR "hip tpls build only supported through spack" )
   endif()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -797,9 +797,12 @@ if( ENABLE_HYPRE )
   endif()
 
   if( ENABLE_HYPRE_DEVICE STREQUAL "CUDA" )
-    list( APPEND HYPRE_DEPENDS chai )
-    string( SUBSTRING ${CUDA_ARCH} 3 -1 HYPRE_CUDA_SM ) # sm_XY -> XY
-    set( HYPRE_CUDA_FLAGS "\
+    if( NOT ENABLE_CUDA )
+      message( WARNING "Setting of HYPRE device to CUDA ignored because ENABLE_CUDA is OFF" )
```
TotoGaz, line 801: Don't you think it would be better to kill the run and let the user adjust its input?

Algiane: Thanks for the quick feedback.

I hesitated between a user warning and an error: I chose the warning because an error is more likely to annoy other users or break existing workflows. Moreover, as an error is "blocking", it leads to more commits / PR (to change the behaviour) than a warning.

Let me know if you want me to turn the warning into an error.

TotoGaz: It's your own `cmake` file so your decision.

Still I think that given the amount of logs our `cmake` script is producing, this warning will go 100% unnoticed.
Therefore people will be fooled.
Also, I do not quite get the "_more commits / PR_" idea. Why would you need any commit to add a `-D...` option?

In the end, the user is asking for incompatible things...
```
In the face of ambiguity, refuse the temptation to guess.
```

Algiane: It should be ready to be merged now.

TotoGaz (approved):

CMakeLists.txt
```diff
@@ -797,9 +797,12 @@ if( ENABLE_HYPRE )
   endif()
 
   if( ENABLE_HYPRE_DEVICE STREQUAL "CUDA" )
-    list( APPEND HYPRE_DEPENDS chai )
-    string( SUBSTRING ${CUDA_ARCH} 3 -1 HYPRE_CUDA_SM ) # sm_XY -> XY
-    set( HYPRE_CUDA_FLAGS "\
+    if( NOT ENABLE_CUDA )
+      message( FATAL_ERROR "Non-compatible options: HYPRE device cannot be setted to CUDA when ENABLE_CUDA is OFF" )
```
CusiniM, line 801:
```suggestion
      message( FATAL_ERROR "Non-compatible options: HYPRE device cannot be set to CUDA when ENABLE_CUDA is OFF" )
```

PR #281, opened 2024-08-28
Set CMP0135 to new

### LINKED ISSUE #277: CMake warnings from ExternalProject_Add
I get tons of warnings from `ExternalProject_Add` calls with CMake 3.26.0. Could set the policy or specify the variable as suggested.

CMake Warning (dev) at /opt/homebrew/Cellar/cmake/3.26.0/share/cmake/Modules/ExternalProject.cmake:3091 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
Call Stack (most recent call first):
  /opt/homebrew/Cellar/cmake/3.26.0/share/cmake/Modules/ExternalProject.cmake:4208 (_ep_add_download_command)
  CMakeLists.txt:203 (ExternalProject_Add)
This warning is for project developers.  Use -Wno-dev to suppress it.

victorapm:
Closes #277

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,9 @@
 cmake_minimum_required(VERSION 3.23.1)
+
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.24.0")
+  cmake_policy(SET CMP0135 NEW) # Set the timestamps of all extracted contents to the time of the extraction
+endif()
+
 ################################
 # GEOS_TPL
 ################################


CMakeLists.txt
```diff
@@ -1,4 +1,9 @@
 cmake_minimum_required(VERSION 3.23.1)
```
rrsettgast, line 1:
```suggestion
cmake_minimum_required(VERSION 3.26.3)
```

@CusiniM @wrtobin I think this is a good version to set...any objection?

rrsettgast (commented):
I think we need a few CMP changes in the main repo as well??

CMakeLists.txt
victorapm: Currently, the newest cmake available on Frontier via module load is `3.23.2`

victorapm:
@rrsettgast @CusiniM should we go with 3.24 also for the TPL repo?

https://github.com/GEOS-DEV/GEOS/blob/41e95e34dfef61b4cf1629e5c9ac425fccad0643/src/CMakeLists.txt#L2

PR #288, opened 2024-11-21
chore: upgrade Trilinos version

CusiniM:
The previous version was giving me some compilation errors with clang16 on macOS so I just upgraded it.


CusiniM:
well, `Epetra`, which we use, is deptrecated so we need to clean up our interface and switch to `Tpetra`. They do have cmake options to still use it but I feel that we should just clean this.

OmarDuran:
> well, `Epetra`, which we use, is deptrecated so we need to clean up our interface and switch to `Tpetra`. They do have cmake options to still use it but I feel that we should just clean this.

Thank you @CusiniM for fixing this. I have tested with apple clang 16.0 and the compilation errors are fixed. However, notice that the action [Rockylinux (8, clang 17.0.6, cuda 12.5)](https://github.com/GEOS-DEV/thirdPartyLibs/actions/runs/11963675215/job/33497971275?pr=288#logs) is failing.

rrsettgast (approved):

PR #289, opened 2024-12-19
Add HIP support

victorapm:
Add HIP build options to RAJA, CHAI, and hypre


rrsettgast:
@victorapm This is not needed anymore?

PR #291, opened 2025-02-06
feat: Add macOS spack environment.

CusiniM:
Good news is that I managed to build through `spack`.

I had to remove a few things that were causing some issues. This is the last build command that worked for me. 

````
python3 scripts/uberenv/uberenv.py --spack-env-file=scripts/spack_configs/macOS/spack.yaml --prefix=/Users/cusini1/geos/geos-tpls --spec "%apple-clang@16.0.0 ~trilinos ~openmp ~scotch ~addr2line"
````
- `~trilinos`: just because we don't really use it anyways and it takes long to build.
- `~openmp`: if this is not specified, the `-fopenmp` passed to the compiler and apple clang does not support it. 
- '~scotch' : this does not build on my mac. It's not a hard dependency but it may require a small patch. 
- `~addr2line`: I am not sure why but a wrong path was added to the resulting host-config for this. I have now specified the correct path in the `spack` environment file but it also does not seem to be a hard requirement so maybe we can add this to the geos spack recipe.  

This PR also upgrades the spack version used so that it supports MacOS sequoia.

### CODE REVIEW

File: scripts/spack_configs/macOS/spack.yaml [hunk 1]
```diff
@@ -0,0 +1,166 @@
...
+      buildable: False
+      externals:
+      - spec: python@3.10.8
+        prefix: /Users/cusini1/pythonVenv/geosxEnv/
+
+    # py-sphinx:
+    #   buildable: False
+    #   externals:
+    #   - spec: sphinx
+    #     prefix: /Users/cusini1/pythonVenv/geosxEnv/
```
CusiniM, line 135:
```suggestion
```

scripts/spack_configs/macOS/spack.yaml
```diff
@@ -0,0 +1,166 @@
...
+      externals:
+        - spec: diffutils@3.11
+          prefix: /opt/homebrew/opt/diffutils
+    # This needs to be the prefix to the pygeosx LC installation
+    # or system install with pygeosx dependencies
+    python:
+      buildable: False
+      externals:
+      - spec: python@3.10.8
+        prefix: /Users/cusini1/pythonVenv/geosxEnv/
```
CusiniM, line 129: Because of things like this, I am not sure how we can write this in a generic way that may work for everybody. Users will need to at least somewhat understand their spack env file.

rrsettgast (approved):

bmhan12 (approved):

PR #305, opened 2025-04-09
refactor: move versions to geos package.py

CusiniM:
Trying to force the definition of the versions of all tpls in the geosx `package.py`. This should ensure that we don't need to specify the commit in all environments and it should make updating a version easier.

### DIFF
--- a/scripts/spack_configs/blueos_3_ppc64le_ib_p9/spack.yaml
+++ b/scripts/spack_configs/blueos_3_ppc64le_ib_p9/spack.yaml
@@ -106,33 +106,6 @@ spack:
     hypre:
       require: "@git.21e5953ddc6daaa24699236108866afa597a415c"
 
-    # v2024.07.0
-    chai:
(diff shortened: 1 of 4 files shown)


victorapm:
Another idea: let's just point to hypre@master. Only drawback is that is a moving target...

CusiniM:
> Another idea: let's just point to hypre@master. Only drawback is that is a moving target...
> 

The thing is that, for Hypre, sometimes we may want commits that are not on master. I also don't know whether that can be specified in the package.py or not. We could try. It's true that ideally we want to control when updates occur. 

> PS: if we are removing all the tarballs, do we want to keep `CMakeLists.txt`?
no, I got rid of it.

CusiniM:
There is one job that fails while building caliper. First of all, I do not understand why but I have also noticed that it's trying to build caliper 2.11 even though the version specified in the geos package is 2.12. I do not understand where it's picking up that version.

victorapm:
Closing since this work has been completed already

PR #306, opened 2025-04-16
fix errors in apple-clang 17

rrsettgast:
- [x] update to VTK 9.4.2
- [x] patch VTK 9.4.2 to avoid compilation errors
- [x] update trilinos to avoid compilation errors

https://github.com/GEOS-DEV/GEOS/pull/3626


scripts/spack_packages/packages/vtk/package.py
```diff
@@ -15,11 +15,16 @@ class Vtk(CMakePackage):
...
     list_url = "http://www.vtk.org/download/"
 
     maintainers = ['chuckatkins', 'danlipsa']
-
+    
+    version(
+        "9.4.2",
+        sha256="36c98e0da96bb12a30fe53708097aa9492e7b66d5c3b366e1c8dc251e2856a02",
+        preferred=True,
+    )
```
CusiniM, line 27: if we need to apply that patch, I think it needs to happen here for spack builds.

CusiniM (approved):

PR #309, opened 2025-05-29
refactor: Consolidate shared package versions for spack.yaml files

bmhan12:
This PR:
- Updates uberenv submodule commit hash for new feature added in LLNL/uberenv#141
- Moves redundant version information into `versions.yaml`, addressing one of the issues discussed in #305, where the same version information does not need to be specified in every `spack.yaml` file.

See GEOS-DEV/GEOS#3697 for testing.

### DIFF
--- a/docker/Stanford/Dockerfile
+++ b/docker/Stanford/Dockerfile
@@ -23,6 +23,7 @@ RUN yum install -y \
     # xz \
     unzip \
     bzip2 \
+    gnupg \
     && pip3 install virtualenv
 
 # Set locale to bypass ascii codec decode error
--- a/docker/TotalEnergies/Dockerfile
+++ b/docker/TotalEnergies/Dockerfile
@@ -22,6 +22,7 @@ RUN yum install --disablerepo=intel-mkl-repo -y \
     # xz \
     unzip \
     bzip2 \
+    gnupg2 \
     && pip3 install virtualenv
 
 # Install clingo for Spack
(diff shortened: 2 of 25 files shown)


rrsettgast:
@bmhan12 When we try to build the TPL's on LC we get the following error:

```
==> Removing install failure marks
==> Removing cached information on repositories
==> Removing python cache files
[creating spack env]
[exe: /usr/WS1/GEOS/GEOSX/TPLs_2025-09-18/dane-gcc-12_tpls/spack/bin/spack env create -d /usr/WS1/GEOS/GEOSX/TPLs_2025-09-18/dane-gcc-12_tpls/dane-gcc-12_env /usr/WS1/geosadmn/geosx/thirdPartyLibs/scripts/spack_configs/toss_4_x86_64_ib/spack.yaml]
==> Error: Detected 2 missing include path(s):
   /usr/WS1/GEOS/GEOSX/TPLs_2025-09-18/dane-gcc-12_tpls/versions.yaml
   /usr/WS1/GEOS/GEOSX/TPLs_2025-09-18/dane-gcc-12_tpls/defaults.yaml
[ERROR: Failed to create Spack Environment]
```

I think this means that the two new yaml files were not copied to the install directory. I.e. they are not present in `/usr/WS1/GEOS/GEOSX/TPLs_2025-09-18/dane-gcc-12_tpls/dane-gcc-12_env`

who should be responsible for copying these? I assume it is the same component that copies `spack.yaml`? And why do the CI jobs work ok?

bmhan12:
> When we try to build the TPL's on LC we get the following error:

This PR does require an update to the uberenv submodule. 
Maybe that might be causing the issue?

rrsettgast:
@bmhan12 well...that is embarrassing

PR #311, opened 2025-06-16
Adding configuration for Dane.

jacobnuttall:
The third party libraries for GEOS are not being compiled on Dane in the LC using the clang compiler. The intent behind this pull request is to add that configuration to the LC installation on Dane.


rrsettgast:
replaced this with #316 which also removes the deceased ruby.

PR #312, opened 2025-06-20
Update hypre

victorapm:
- Add physics-based scaling (host + device)
- MGR performance improvements (device)
- Add tagged residual capabilities for analyzing convergence history for different physics DOFs
- CMake build fixes

https://github.com/GEOS-DEV/GEOS/pull/3682

### DIFF
--- a/scripts/updateHypre.sh
+++ b/scripts/updateHypre.sh
@@ -48,21 +48,21 @@ rm -rf ${HYPRE_DIR}/src/test/TEST_*
 rm -rf ${HYPRE_DIR}/.git
 
 # Remove old hypre dir
-git rm ${TPL_MIRROR_DIR}/hypre*.tar.gz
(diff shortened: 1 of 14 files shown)


scripts/updateHypre.sh
```diff
@@ -48,21 +48,21 @@ rm -rf ${HYPRE_DIR}/src/test/TEST_*
 rm -rf ${HYPRE_DIR}/.git
 
 # Remove old hypre dir
-git rm ${TPL_MIRROR_DIR}/hypre*.tar.gz
+#git rm ${TPL_MIRROR_DIR}/hypre*.tar.gz
```
rrsettgast, line 51: @victorapm 
what is going on with this file?

victorapm: This was my old script to retrieve hypre hash, download and pack the tarball, but we don't need all that anymore. So I commented out the parts we don't need

PR #314, opened 2025-08-06
Update scotch to 7.0.8 and add/fix determinism flags

klevzoff:
Looks like Scotch config in https://github.com/GEOS-DEV/thirdPartyLibs/pull/310 resulted in some non-deterministic behavior which causes one integrated test to constantly fail baseline checks in https://github.com/GEOS-DEV/GEOS/pull/2427. This was partly due to a bug in Scotch CMake that was recently [fixed](https://gitlab.inria.fr/scotch/scotch/-/commit/ac56fdbac406b92f8595d7cbd88d1e387d4ec2e3). To get fully reproducible behavior between runs, one must configure with `SCOTCH_DETERMINISTIC=FULL` cmake flag.

This PR updates Scotch to latest v7.0.8 that includes the fix and updates Spack package recipes to configure Scotch correctly. Note that this fix does not yet exist in upstream Spack, so my approach may differ from what package maintainers eventually decide on.

Once this is merged, https://github.com/GEOS-DEV/GEOS/pull/2427 will be updated to a new `GEOS_TPL_TAG`.

### DIFF
--- a/docker/Stanford/Dockerfile
+++ b/docker/Stanford/Dockerfile
@@ -49,8 +49,6 @@ RUN yum install -y \
     make \
     bc \
     file \
-    bison \
-    flex \
     patch \
     ca-certificates \
     autoconf-2.69 \
--- a/docker/TotalEnergies/Dockerfile
+++ b/docker/TotalEnergies/Dockerfile
@@ -38,8 +38,6 @@ RUN yum install --disablerepo=intel-mkl-repo -y \
     make \
     bc \
     file \
-    bison \
-    flex \
     patch \
     ca-certificates \
     autoconf \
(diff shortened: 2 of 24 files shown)


klevzoff:
Had some problems with system-installed `flex` version on Rocky linux, so went nuclear in last two commits and removed system installed flex/bison from all images :sweat_smile: Scotch is quite finicky when it comes to flex versions, so trying to just let spack install the right version. Could've done this just for two problematic images instead, but let's see if this works, might save some future headaches.

rrsettgast (approved):

OmarDuran:
> Had some problems with system-installed `flex` version on Rocky linux, so went nuclear in last two commits and removed system installed flex/bison from all images 😅 Scotch is quite finicky when it comes to flex versions, so trying to just let spack install the right version. Could've done this just for two problematic images instead, but let's see if this works, might save some future headaches.

I had similar issues with system-installed flex version on macOS. These changes make a lot of sense to me.

PR #324, opened 2025-10-02
[WIP] Spack 1.0.2

bmhan12:
This PR:
- Attempts to move spack version to 1.0.2


victorapm:
@matteofrigo5 we're having issues with the Sherlock jobs and it's not very clear what's going on.

Given it's using centos 7, which already reached EOL, how about we upgrade those images to let's say centos 9? Would that work for Sherlock?

PR #351, opened 2026-05-18
move to c++20

rrsettgast:
- [x] update the fmt spec in spack to 12.1.0 to avoid compiler errors when compiling with c++20.
- [ ] confirm geos compiles once it moves to c++20

### DIFF
--- a/scripts/spack_packages/packages/geosx/package.py
+++ b/scripts/spack_packages/packages/geosx/package.py
@@ -77,7 +77,8 @@ class Geosx(CMakePackage, CudaPackage, ROCmPackage):
             multi=False)
     variant('grpc', default=False, description='Enable gRPC.')
     variant('pygeosx', default=True, description='Enable pygeosx.')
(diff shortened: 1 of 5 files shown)


scripts/spack_packages/packages/geosx/package.py
```diff
@@ -93,14 +94,17 @@ class Geosx(CMakePackage, CudaPackage, ROCmPackage):
     variant('cuda_stack_size', default="0", description="Defines the adjusted cuda stack \
         size limit if required. Zero or negative keep default behavior")
 
+    requires('cxxstd=20', msg='GEOS.AI TPL snapshots require C++20.')
```
rrsettgast:
```suggestion
    requires('cxxstd=20', msg='GEOS TPL snapshots require C++20.')
```

victorapm:
This will be interesting for the device builds as well because the latest umpire release has moved to `C++20`

victorapm:
Transition to C++20 already completed

PR #359, opened 2026-08-05
Update spack v1.2.2

victorapm:
Update spack to `v1.2.2`, which supports a new parallel installer for faster builds.

In addition, this PR removes old code for building on retired machines


victorapm:
@rrsettgast @joshua-white This is ready for review and merge

rrsettgast (approved):

rrsettgast:
@victorapm is this built on LC and is there a corresponding GEOS PR?

victorapm:
Thanks Randy, I didn't attach a GEOS PR here and this hasn't been built on LC. I thought since this is only bumping the spack commit and doing some cleanup (no TPL changes here), an LC build wasn't needed. Let me know if we should do it

